PR #2, opened 2021-06-24
Add .gitignore

johann-beleites-sonarsource:


.gitignore
```diff
@@ -0,0 +1,42 @@
+# --- Gradle
+.gradle/
+build/
+
+# ---- Maven
+target/
+dependency-reduced-pom.xml
```
Godin, line 7: I don't think that we plan to go back to Maven, so propose to remove this and in general have only really necessary things in `.gitignore`, as a motivational example - presence of `target/` in `.gitignore` means that on attempt to add package named `target` you might forget to force-add it :wink:

Godin (changes requested):

PR #3, opened 2021-06-24
SONARKT-77 Removing vast majority of remaining SLang API usages

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/testing/AbstractSensorTest.kt
```diff
@@ -0,0 +1,88 @@
...
+            .setModuleBaseDir(baseDir.toPath())
+            .setType(InputFile.Type.MAIN)
+            .setLanguage(language().key)
+            .setCharset(StandardCharsets.UTF_8)
+            .setContents(content)
+            .build()
+    }
+
+    protected abstract fun repositoryKey(): String
+    protected abstract fun language(): Language
```
margarita-nedzelska-sonarsource, line 87: It looks like we don't need these 2 methods being abstract, because the implementation is always the same

sonar-kotlin-plugin/src/test/java/org/sonar/api/utils/log/ThreadLocalLogTester.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public ThreadLocalLogTester clear() {
+    INTERCEPTOR.clear();
+    return this;
+  }
+
+  private static final class Interceptor implements LogInterceptor {
+
+    private static final ThreadLocal<List<String>> LOGS = ThreadLocal.withInitial(ArrayList::new);
```
margarita-nedzelska-sonarsource: It's definitely not for now. But I'd suggest we rethink this logic and use coroutines context instead of using ThreadLocal because I doubt that we really need a ThreadLocal here

margarita-nedzelska-sonarsource (commented):
Nice work!
There are a few minor comments to improve.

I'm not approving at the moment, because I would like to have a second path through it before merging. Just in case I missed something.

And anyway I think we can refactor it in the future, and get rid of the rest of slang-like code in the future

sonar-kotlin-plugin/src/test/java/org/sonar/api/utils/log/ThreadLocalLogTester.java
johann-beleites-sonarsource: Yea, I agree. Not very Kotlin-like :smile:

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CommentedCodeCheck.kt
```diff
@@ -28,8 +28,9 @@ import org.sonar.check.Rule
 import org.sonarsource.kotlin.api.AbstractCheck
 import org.sonarsource.kotlin.api.getContent
 import org.sonarsource.kotlin.converter.KotlinCodeVerifier
+import org.sonarsource.kotlin.converter.KotlinTextRanges.merge
+import org.sonarsource.kotlin.converter.KotlinTextRanges.textRange
 import org.sonarsource.kotlin.plugin.KotlinFileContext
-import org.sonarsource.slang.impl.TextRanges
 
 /**
  * Replacement for [org.sonarsource.slang.checks.CommentedCodeCheck]
```
margarita-nedzelska-sonarsource, line 36: We can get rid of such comments (in all the checks), as these classes do not exist anymore in this repository and we can modify the implementation in the future.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/DuplicateBranchCheck.kt
```diff
@@ -39,9 +38,9 @@ class DuplicateBranchCheck : AbstractBranchDuplication() {
             val original = group[0]
             group.asSequence()
                 .drop(1)
-                .filter { spansMultipleLines(it, document) }
```
margarita-nedzelska-sonarsource, line 42: Looks like the variable on line 36 is not used anymore

margarita-nedzelska-sonarsource (approved):
Looks good!

Just few minor comments.

PR #7, opened 2021-06-29
SONARKT-76 Improve and unify the check testing API

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/CheckTest.kt
```diff
@@ -0,0 +1,54 @@
...
+            this@CheckTest.dependencies?.let { this.deps = it }
+        }.verify()
+    }
+
+    @Test
+    fun `without semantics`() {
+        assumeTrue(shouldRunTestWithoutSemantics())
+
+        KotlinVerifier(check) {
+            this.fileName = sampleFileNoSemantics ?: "${checkName}SampleNoSemantics.kt"
```
margarita-nedzelska-sonarsource, line 44: maybe you can reuse the constant NO_SEMANTICS_TEST_FILE_POSTFIX here?

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/CheckTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.Assumptions.assumeTrue
+import org.junit.jupiter.api.Test
+import org.sonarsource.kotlin.api.AbstractCheck
+import org.sonarsource.kotlin.verifier.KotlinVerifier
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.io.path.isRegularFile
+import kotlin.streams.asSequence
+
+private val checkNamesWithAvailableSampleTestFiles by lazy {
```
margarita-nedzelska-sonarsource, line 12: by default `lazy` does synchronization and it might affect the performance. So it's better to use `lazy(LazyThreadSafetyMode.NONE)`

margarita-nedzelska-sonarsource (commented):
Overall looks good.

I just have a small concern regarding re-running or debugging single tests.
What if I need to debug only `without semantic` check. I can only edit my configuration in IDE, but cannot easily do it from IDE.

It looks like currently, we have only 2 no semantic tests. Maybe it will be more convenient to write them explicitly in the test class and keep only the `with semantic` test in CheckTest?

And I feel that 50 additional `skipped tests` pollute the log

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/TabsCheckTest.kt
```diff
@@ -19,23 +19,4 @@
...
-
-    @Test
-    fun `no issues`() {
-        KotlinVerifier(TabsCheck()) {
-            fileName = "DummyNoIssues.kt"
-        }.verifyNoIssue()
-    }
-
-}
+class TabsCheckTest : CheckTestWithNoSemantics(TabsCheck(), sampleFileNoSemantics = "DummyNoIssues.kt")
```
margarita-nedzelska-sonarsource, line 22: Sorry, just noticed. In this test "DummyNoIssues.kt" file is not used to test the check without semantic.
This rule doesn't care about semantic at all.

The second test is a "happy path", to test when the issue is not raised. This cannot be tested using one file, because this rule raises an issue on the whole file.

While technically using CheckTestWithNoSemantics here and having a test without semantic is not harmful, it's a little bit confusing, because the purpose is different

margarita-nedzelska-sonarsource (approved):
One small comment, other than that looks good!

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/TabsCheckTest.kt
johann-beleites-sonarsource: Ah okay I see, no problem, I'll just add the test separately

PR #8, opened 2021-06-29
Migrate main build file from Gradle Groovy DSL to Kotlin DSL

johann-beleites-sonarsource:


johann-beleites-sonarsource:
Depends on #34

Godin (approved):

PR #10, opened 2021-06-30
Bugfix: setting a FunMatcher's name after the matcher has already bee…

johann-beleites-sonarsource:
…n instantiated is now actually applied.


margarita-nedzelska-sonarsource (approved):
Nice spot!

PR #12, opened 2021-06-30
SONARKT-50 Rule s6288 Authorizing non-authenticated users to use keys…

margarita-nedzelska-sonarsource:
… in the Android KeyStore is security-sensitive


johann-beleites-sonarsource (approved):
Looks good, LGTM!

PR #14, opened 2021-07-07
Upgrade Gradle to 7.1.1

Godin:


margarita-nedzelska-sonarsource (commented):
Do you think it makes sense to bump the shadow plugin to the 7th version in the same PR?

margarita-nedzelska-sonarsource (approved):

PR #16, opened 2021-07-14
SONARKT-94 S6318: "suspend" modifier should not be redundant

margarita-nedzelska-sonarsource:


johann-beleites-sonarsource (approved):
Nice, LGTM!

PR #18, opened 2021-07-15
SONARKT-93 S6315: The 'Deferred' return value should be used

margarita-nedzelska-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnusedAsyncResultCheck.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.sonarsource.kotlin.api.AbstractCheck
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+@Rule(key = "S6315")
+class UnusedAsyncResultCheck: AbstractCheck() {
+
+    override fun visitCallExpression(expression: KtCallExpression, context: KotlinFileContext) {
+        val bindingContext = context.bindingContext
+        if (bindingContext.get(BindingContext.EXPRESSION_TYPE_INFO, expression)?.type?.getJetTypeFqName(false) != "kotlinx.coroutines.Deferred") return
+        if (expression.isUsedAsStatement(bindingContext)) {
```
johann-beleites-sonarsource, line 17: Effectively these 2 if-blocks are a `&&` - I think it could make sense to write it as one as well, to be a bit more explicit about it

johann-beleites-sonarsource (approved):
I have one suggestion, otherwise LGTM!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnusedDeferredResultCheck.kt
```diff
@@ -0,0 +1,27 @@
...
+        if (expression.expressionTypeFqn(bindingContext) == DEFERRED_FQN
+            && expression.isUsedAsStatement(bindingContext)
+        ) {
+            context.reportIssue(expression.calleeExpression!!, """This function returns "Deferred", but its result is never used""")
+            return
+        }
+    }
+
+    private fun KtCallExpression.expressionTypeFqn(bindingContext: BindingContext) =
+        bindingContext.get(BindingContext.EXPRESSION_TYPE_INFO, this)?.type?.getJetTypeFqName(false)
```
johann-beleites-sonarsource, line 26: Actually, might this make sense in the API extensions?

PR #20, opened 2021-07-20
SONARKT-99 S6305: "MutableXFlow" not be exposed

johann-beleites-sonarsource:


margarita-nedzelska-sonarsource (approved):
Other than that looks good 👍

PR #21, opened 2021-07-21
SONARKT-98 S6309: Functions returning Flow/Channel should not be suspending

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ApiExtensions.kt
```diff
@@ -252,3 +253,9 @@ private fun KtParameter.determineType(bindingContext: BindingContext) =
 
 fun KtParameter.determineTypeAsString(bindingContext: BindingContext, printTypeArguments: Boolean = false) =
     determineType(bindingContext)?.getJetTypeFqName(printTypeArguments)
+
+fun KtNamedFunction.returnTypeAsString(bindingContext: BindingContext, printTypeArguments: Boolean = false) =
+    returnType(bindingContext)?.getJetTypeFqName(printTypeArguments)
+
+fun KtNamedFunction.returnType(bindingContext: BindingContext) =
```
margarita-nedzelska-sonarsource, line 260: It looks like this function could be private. Also. maybe it makes sense to combine this function with `returnTypeAsString`?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/FlowChannelReturningFunsNotSuspendingCheck.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+private val FORBIDDEN_RETURN_TYPES = listOf(COROUTINES_FLOW, COROUTINES_CHANNEL)
+private const val MESSAGE = """Functions returning "Flow" or "Channel" should not be suspending"""
+
+@Rule(key = "S6309")
+class FlowChannelReturningFunsNotSuspendingCheck : AbstractCheck() {
+    override fun visitNamedFunction(function: KtNamedFunction, kotlinFileContext: KotlinFileContext) {
+        val suspendModifier = function.suspendModifier()
+        if (suspendModifier != null && function.returnTypeAsString(kotlinFileContext.bindingContext) in FORBIDDEN_RETURN_TYPES) {
```
margarita-nedzelska-sonarsource, line 21: How about `function.suspendModifier()?.let {...}`?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/FlowChannelReturningFunsNotSuspendingCheck.kt
```diff
@@ -0,0 +1,29 @@
...
+            val secondaries = function.typeReference
+                ?.let { SecondaryLocation(kotlinFileContext.textRange(it)) }
+                ?.let { listOf(it) }
```
margarita-nedzelska-sonarsource, line 24:
```suggestion
                ?.let { listOf(SecondaryLocation(kotlinFileContext.textRange(it))) }
```

margarita-nedzelska-sonarsource (approved):
Overall looks good, few minor suggestions

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ApiExtensions.kt
johann-beleites-sonarsource: My idea behind this was that you may want to perform some further processing on the type, for which you might need the actual `KotlinType`. So by splitting it up like this, you still have that option in the API. WDYT?

PR #23, opened 2021-07-22
SONARKT-97 S6312: Extension functions on CoroutineScope should not be suspending

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CoroutineScopeFunSuspendingCheck.kt
```diff
@@ -0,0 +1,36 @@
...
+private const val MESSAGE = "Extension functions on CoroutineScope should not be suspending"
+
+@Rule(key = "S6312")
+class CoroutineScopeFunSuspendingCheck : AbstractCheck() {
+    override fun visitNamedFunction(function: KtNamedFunction, kotlinFileContext: KotlinFileContext) {
+        // Only applicable for extension functions
+        if (!function.isExtensionDeclaration()) return
+
+        val suspendModifier = function.suspendModifier() ?: return
+        val receiverType = function.receiverTypeReference ?: return
```
margarita-nedzelska-sonarsource, line 25: It looks like this condition cannot fail, because of line L22. You can simplify it a little bit and remove the check on L22 at all

margarita-nedzelska-sonarsource (approved):
Looks good!

PR #25, opened 2021-07-23
Improvements to the FunMatcher API

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -75,9 +81,12 @@ class FunMatcher(
     private fun checkFunctionDescriptor(functionDescriptor: CallableDescriptor?) =
         functionDescriptor != null &&
             checkIsDynamic(functionDescriptor) &&
+            checkIsExtensionFunction(functionDescriptor) &&
+            checkIsSuspending(functionDescriptor) &&
             checkName(functionDescriptor) &&
             checkTypeOrSupertype(functionDescriptor) &&
-            checkCallParameters(functionDescriptor)
+            checkCallParameters(functionDescriptor) &&
+            checkReturnType(functionDescriptor)
```
margarita-nedzelska-sonarsource, line 89: I'd rather swapped this 2 calls, because I feel `checkReturnType` is faster, so there is no need to check parameters, if return type check fails.

margarita-nedzelska-sonarsource (approved):
Overall looks good, just one small point

PR #26, opened 2021-07-23
SONARKT-96 S6314: Intermediate Flow operations should be used

johann-beleites-sonarsource:


kotlin-checks-test-sources/src/main/kotlin/checks/FinalFlowOperationCheckSample.kt
```diff
@@ -0,0 +1,26 @@
+package checks
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.count
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
+
+class FinalFlowOperationCheckSample {
+    suspend fun foo(flow: Flow<String>) {
```
margarita-nedzelska-sonarsource, line 10: I think it will be nice to add a few more cases, where `Flow` is returned as a result of a lambda or passed as an argument, or returned from a function, etc. They should be handled by the implementation, but still nice to double-check.

margarita-nedzelska-sonarsource (commented):
Looks good!

Not approving yet, in case after adding new cases,  you find anything strange (probably should not)

kotlin-checks-test-sources/src/main/kotlin/checks/FinalFlowOperationCheckSample.kt
johann-beleites-sonarsource: WDYT of these new ones?

margarita-nedzelska-sonarsource: Looks good, I've just realized you're not checking the message and precise issue location.

margarita-nedzelska-sonarsource (approved):
Looks good, just add message and issue location in tests in one place to be sure everything is tested.

PR #27, opened 2021-07-26
SONARKT-95 S6316: Kotlin coroutines API for timeouts should be used 

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CoroutinesTimeoutApiUnusedCheck.kt
```diff
@@ -0,0 +1,76 @@
...
+    override fun functionsToVisit() = listOf(FunMatcher(supertype = "$KOTLINX_COROUTINES_PACKAGE.Job", name = "cancel"))
+
+    override fun visitFunctionCall(
+        cancelCallExpr: KtCallExpression,
+        resolvedCancelCall: ResolvedCall<*>,
+        kotlinFileContext: KotlinFileContext
+    ) {
+        val bindingContext = kotlinFileContext.bindingContext
+
+        val cancelCallCalleeExpression = cancelCallExpr.calleeExpression ?: return
```
margarita-nedzelska-sonarsource, line 39: Not obligatory, but according to my experience: It seems that if you've already matched some function signature, `caleeExpression` won't be null, so you can use non-null assertion safely.

kotlin-checks-test-sources/src/main/kotlin/checks/CoroutinesTimeoutApiUnusedCheckSample.kt
```diff
@@ -0,0 +1,101 @@
...
+            }
+            delay(500L)
+//          ^^^^^>
+            job.cancel()
+//              ^^^^^^
+        }
+
+        coroutineScope {
+            println("")
+            val job = async() {
```
margarita-nedzelska-sonarsource, line 29: Looks like it's better to name the result `deferred` and add examples where result of `async` function is actually used (with relevant compliant example) to make sure we're not suggesting anything invalid.

For example:

```
        coroutineScope {
            val deferred = async() {
                delay(2000L)
                println("Finished")
                100
            }
            delay(500L)
            deferred.cancel() // Noncompliant
            deferred.await()
        }

        coroutineScope {
            println("")
            val i = withTimeoutOrNull(500L) {
                delay(2000L)
                println("Finished")
                100
            }
            i
        }
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CoroutinesTimeoutApiUnusedCheck.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        val jobDeclaration = (cancelCallExpr.context as? KtDotQualifiedExpression)
+            ?.receiverExpression?.getReferenceTargets(bindingContext)?.toList()?.getOrNull(0) ?: return
+
+        val siblingSequence = cancelCallExpr.parent.siblings(forward = false, withItself = false)
+            .filter { it is KtElement }
+
+        var delayCall: KtExpression? = null
+        var initializerCall: KtExpression? = null
+        for (sibling in siblingSequence) {
```
margarita-nedzelska-sonarsource, line 49: How about moving code inside `if` and `else` to a separate methods and use sth as `sequence.iterator().next()` instead of for loop?

Despite this for loop serves its purpose, it was not quite obvious for me to understand what it actually does and why it works.

margarita-nedzelska-sonarsource (approved):
Overall looks good, just a few comments

PR #29, opened 2021-07-28
SONARKT-91 S6311: Suspending functions should be called on the caller…

margarita-nedzelska-sonarsource:
…'s thread


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SuspendingFunCallerDispatcherCheck.kt
```diff
@@ -0,0 +1,38 @@
...
+    override fun visitFunctionCall(
+        callExpression: KtCallExpression,
+        resolvedCall: ResolvedCall<*>,
+        kotlinFileContext: KotlinFileContext,
+    ) {
+        val bindingContext = kotlinFileContext.bindingContext
+
+        val arguments = resolvedCall.valueArgumentsByIndex
+        if (arguments == null || arguments.isEmpty()) return
+        if (arguments[0] == null || arguments[0] is DefaultValueArgument) return
```
johann-beleites-sonarsource, line 27: Any reason not to reduce this to 1 `if` statement?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SuspendingFunCallerDispatcherCheck.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        /* Last Call Expression is always the call itself, so we drop it */
+        val callExpressions = callExpression.collectDescendantsOfType<KtCallExpression>().dropLast(1)
+        if (callExpressions.isNotEmpty()
+            && callExpressions.all { it.getResolvedCallNative(bindingContext)?.resultingDescriptor?.isSuspend == true }
+        ) {
+            kotlinFileContext.reportIssue(callExpression.calleeExpression!!,
```
johann-beleites-sonarsource, line 34: How about raising the issue on the dispatcher instead of the function name? That would fit a bit better to the message saying "Remove this dispatcher..."

johann-beleites-sonarsource (approved):
A couple of suggestions, then LGTM!

PR #30, opened 2021-07-28
SONARKT-80 Update Kotlin to 1.5.21

Godin:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/WeakSSLContextCheck.kt
```diff
@@ -134,7 +134,7 @@ class WeakSSLContextCheck : AbstractCheck() {
         is KtStringTemplateExpression -> asConstant()
         is KtNameReferenceExpression -> {
             val descriptor = bindingContext.get(REFERENCE_TARGET, this)
-            if (descriptor is PropertyDescriptor) descriptor.compileTimeInitializer?.stringTemplateValue()
```
margarita-nedzelska-sonarsource, line 137: It looks like `toString()` method has exactly the same implementation as `stringTemplateValue()` and combination of `boxedValue().toString()` does the same but is just longer.

margarita-nedzelska-sonarsource (approved):
Overall looks good, just 1 minor comment!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/WeakSSLContextCheck.kt
Godin: Already [in Kotlin 1.5.10 `ConstantValue.toString` was doing the same thing as `ConstantValue.stringTemplateValue`](https://github.com/JetBrains/kotlin/blob/v1.5.10/core/descriptors/src/org/jetbrains/kotlin/resolve/constants/constantValues.kt#L44-L46), yet we were using `stringTemplateValue`, which IMO makes sense from clarity point of view and IMO same now -
`ConstantValue.boxedValue().toString()` is more clear than `ConstantValue.toString()`

PR #37, opened 2021-07-28
Fix Gradle warning about deprecated property JavaExec.main

Godin:
See https://docs.gradle.org/7.1.1/dsl/org.gradle.api.tasks.JavaExec.html#org.gradle.api.tasks.JavaExec:main


johann-beleites-sonarsource (approved):
LGTM!

PR #42, opened 2021-07-29
SONARKT-82 Fix false-positives with backticks in S1172 UnusedFunction…

margarita-nedzelska-sonarsource:
…ParameterCheck

PR #44, opened 2021-07-29
SONARKT-83 Fix false-negatives in S1172 UnusedFunctionParameterCheck

margarita-nedzelska-sonarsource:
* anonymous functions
* lambdas


margarita-nedzelska-sonarsource:
Should be merged after https://github.com/SonarSource/sonar-kotlin/pull/42. I'll handle the conflicts

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnusedFunctionParameterCheck.kt
```diff
@@ -19,54 +19,84 @@
...
             }
             .toList()
         val firstUnused = unusedParameters[0]
-        val msg = if (unusedParameters.size > 1) {
-            "Remove these unused function parameters."
-        } else {
-            "Remove this unused function parameter \"${firstUnused.name}\"."
-        }
+        val msg = if (unusedParameters.size > 1) pluralMessage
+        else """$singleMessage "${firstUnused.name}"."""
```
johann-beleites-sonarsource, line 73: Currently, we raise the same message for the same location twice when we have only 1 unused parameter. I don't think that's valuable and it may even look like a bug, as the UI shows the same thing twice. Can we get rid of the secondary location in case there is only 1 unused parameter?

johann-beleites-sonarsource (approved):
One suggestion, otherwise looks great!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnusedFunctionParameterCheck.kt
margarita-nedzelska-sonarsource: Yep, I was thinking about it. We decided to leave it previously, to be consistent with what we had in SLang, but currently, it makes sense to fix this.

PR #49, opened 2021-08-05
SONARKT-48 Rule S6301: Mobile database encryption keys should not be …

margarita-nedzelska-sonarsource:
…disclosed


kotlin-checks-test-sources/src/main/kotlin/checks/MobileDatabaseEncryptionKeysCheckSample.kt
```diff
@@ -0,0 +1,91 @@
+package checks
+
+import io.realm.RealmConfiguration
+import net.sqlcipher.database.SQLiteDatabase
+
+class MobileDatabaseEncryptionKeysCheckSample(val k: Key) {
+
+    fun realm(s: String) {
+        val stringKey = "d2345678e012345678901c3456789012a456789012n45678901234o678901227" // Secondary location here
```
johann-beleites-sonarsource, line 9: How about verifying that the secondary location is really marked? You could also do this by constructing a smaller example in a new method to avoid having a `// Noncompliant@+veryLargeNumber` comment.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MobileDatabaseEncryptionKeysCheck.kt
```diff
@@ -0,0 +1,94 @@
...
+
+@Rule(key = "S6301")
+class MobileDatabaseEncryptionKeysCheck : CallAbstractCheck() {
+    override fun functionsToVisit() = listOf(
+        ConstructorMatcher(typeName = SQLITE),
+        FunMatcher(qualifier = SQLITE) {
+            withNames(CHANGE_PASSWORD, "openDatabase", "openOrCreateDatabase", "create")
+        },
+        FunMatcher(qualifier = "io.realm.RealmConfiguration.Builder", name = ENCRYPTION_KEY)
+    )
```
johann-beleites-sonarsource, line 38: We call `functionsToVisit()` a lot, every time we visit a call expression. So every time we need to build the list containing the newly constructed matchers. WDYT of making the list static and returning the same instance every time?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MobileDatabaseEncryptionKeysCheck.kt
```diff
@@ -0,0 +1,94 @@
...
+            kotlinFileContext.reportIssue(
+                callExpression.calleeExpression!!,
+                """The "$parameter" parameter should not be hardcoded.""",
+                secondaries.map { SecondaryLocation(kotlinFileContext.textRange(it)) },
+            )
+        }
+    }
+}
+
+private fun KtExpression.isHardCoded(bindingContext: BindingContext, secondaries: MutableList<PsiElement>) : Boolean =
```
johann-beleites-sonarsource, line 70: Do expressions in parentheses work here? So e.g. `val foo = ("hardcoded")`? What about `val foo = (((((((("hardcoded"))))))))`?

kotlin-checks-test-sources/src/main/kotlin/checks/MobileDatabaseEncryptionKeysCheckSample.kt
johann-beleites-sonarsource, line 9: How about also testing values in parentheses? `val stringKey2 = ((("abc")))`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MobileDatabaseEncryptionKeysCheck.kt
```diff
@@ -0,0 +1,94 @@
...
+    val declaration = DescriptorToSourceUtils.descriptorToDeclaration(resultingDescriptor) as? KtNamedFunction ?: return false
+
+    if (!declaration.hasBody()) return false
+    if (declaration.hasBlockBody()) {
+        return declaration.anyDescendantOfType<KtReturnExpression> {
+            it.returnedExpression?.isHardCoded(bindingContext, secondaries) ?: false
+        }
+    }
+
+    return declaration.bodyExpression?.isHardCoded(bindingContext, secondaries) ?: false
```
johann-beleites-sonarsource, line 93: To me this seems like a logical `else` case for the `if(declaration.hasBlockBody())` statement

johann-beleites-sonarsource (commented):
Looks good overall, I've got a couple of comments

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MobileDatabaseEncryptionKeysCheck.kt
margarita-nedzelska-sonarsource: How about making this as property instead of a function in the CallAbstractCheck

johann-beleites-sonarsource: I think that's a good idea :)

johann-beleites-sonarsource (approved):
LGTM!

PR #53, opened 2021-08-06
Improve messages 

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
As discussed with @Godin, messages ending with regular expressions should not end with dot and not have quotes, to avoid misinterpretation of such symbols as parts of regex.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/BadClassNameCheck.kt
```diff
@@ -51,7 +51,7 @@ class BadClassNameCheck : AbstractCheck() {
         if (ktClass !is KtEnumEntry && !name.matches(formatRegex)) {
             context.reportIssue(
                 ktClass.nameIdentifier!!,
-                "Rename class \"$name\" to match the regular expression $format.",
+                """Rename class "$name" to match the regular expression: $format""",
```
johann-beleites-sonarsource, line 54: I'm not convinced. Why are we removing the `.` here? Because it could belong to the regex? In that case I would suggest using `"[regex]"` instead. 
Also, I think there should be no colon (`:`). The regex belongs to `the regular expression`, it's not something that provides a specification on something more generic (e.g. something like `...to match a valid format: regex` would be fine). Here, though, the regex is part of the regular sentence structure, so no `:` is needed.
```suggestion
                """Rename class "$name" to match the regular expression "$format".""",
```
EDIT: I saw that we already do this anyway in `VariableAndParameterNameCheck`.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/BadFunctionNameCheck.kt
```diff
@@ -52,7 +52,8 @@ class BadFunctionNameCheck : AbstractCheck() {
         if (!name.matches(formatRegex)) {
             kotlinFileContext.reportIssue(
                 function.nameIdentifier!!,
-                "Rename function \"$name\" to match the regular expression $format")
+                """Rename function "$name" to match the regular expression: $format"""
```
johann-beleites-sonarsource, line 55: Same as for `BadClassNameCheck`, I suggest removing `:` and perhaps adding `"$format".` here.
```suggestion
                """Rename function "$name" to match the regular expression "$format"."""
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnusedLocalVariableCheck.kt
```diff
@@ -58,7 +58,8 @@ class UnusedLocalVariableCheck : AbstractCheck() {
                         reference -> reference.getReferencedName() == property.name
             }) {
                 context.reportIssue(nameIdentifier,
-                    "Remove this unused \"${nameIdentifier.text}\" local variable.")
+                    """Remove this unused "${nameIdentifier.text}" local variable."""
+                )
```
johann-beleites-sonarsource, line 62: Can we put it on 1 line, if we're touching it anyway?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CodeAfterJumpCheck.kt
```diff
@@ -41,7 +41,8 @@ class CodeAfterJumpCheck : AbstractCheck() {
             val current = statementsOrExpressions[i]
             getJump(current)?.let {
                 context.reportIssue(current,
-                    "Refactor this piece of code to not have any dead code after this \"$it\".")
+                    """Refactor this piece of code to not have any dead code after this "$it"."""
```
johann-beleites-sonarsource, line 44: We could put this on one line

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/BiometricAuthWithoutCryptoCheck.kt
```diff
@@ -36,7 +36,7 @@ private val AUTHENTICATE_FUN_MATCHERS = listOf(
     FunMatcher(qualifier = "androidx.biometric.BiometricPrompt", name = "authenticate")
 )
 
-private const val MESSAGE = "Make sure performing a biometric authentication without a CryptoObject is safe here"
+private const val MESSAGE = "Make sure performing a biometric authentication without a CryptoObject is safe here."
```
johann-beleites-sonarsource, line 39: Do we need `""` around the `CryptoObject` here?
```suggestion
private const val MESSAGE = "Make sure performing a biometric authentication without a "CryptoObject" is safe here."
```

kotlin-checks-test-sources/src/main/kotlin/checks/UnusedDeferredResultCheckSample.kt
```diff
@@ -9,28 +9,28 @@ import kotlinx.coroutines.delay
 
 suspend fun unused() {
     
-    coroutineScope { // Noncompliant {{This function returns "Deferred", but its result is never used}} 
+    coroutineScope { // Noncompliant {{This function returns "Deferred", but its result is never used.}}
```
johann-beleites-sonarsource, line 12: Can we remove all locations checking the message except for the first one? The message here is always the same, so I see no value in asserting it every time - it just creates more work when changing it.

johann-beleites-sonarsource (commented):
Thanks for having a look at all the messages, looks good! I'm going to nit-pick a little here and there, while were at it :slightly_smiling_face: .

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/VariableAndParameterNameCheck.kt
```diff
@@ -77,7 +77,8 @@ class VariableAndParameterNameCheck : AbstractCheck() {
         if (!name.matches(formatRegex)) {
             kotlinFileContext.reportIssue(
                 declaration.nameIdentifier!!,
-                "Rename this $kind to match the regular expression \"$format\".")
+                """Rename this $kind to match the regular expression "$format"."""
```
johann-beleites-sonarsource, line 80: Remove quotes & dot from regex. (And if you decide to put/leave the colons on the other rules, put one here as well).

johann-beleites-sonarsource (approved):
Actually, all suggestions are pretty minor, so unless there is the need to discuss anything, go ahead and merge afterwards.

PR #61, opened 2021-08-10
SONARKT-107 Rule S1874: Deprecated code should not be used

Godin:


margarita-nedzelska-sonarsource:
I just realized this check might be tricky, because we are always using the latest Kotlin version to generate the environment and some elements could be deprecated in the latest version but not deprecated in the version used by the user. So this will result in False positives.

margarita-nedzelska-sonarsource:
FYI: I checked that Kotlin stdlib version in the classpath is set correctly from the Gradle configuration if run with the Gradle scanner, so we can report this diagnistic without any fears.

margarita-nedzelska-sonarsource (approved):

PR #62, opened 2021-08-11
Upgrading plugin API & removing JUnit 4

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/dev/AstPrinter.kt
```diff
@@ -107,3 +105,18 @@ fun textPointerAtOffset(psiDocument: Document, startOffset: Int): TextPointer {
 
     return DefaultTextPointer(startLineNumber + 1, startLineOffset)
 }
+
+private class DefaultTextPointer(val line: Int, val lineOffset: Int) : TextPointer {
+    override fun line() = line
+    override fun lineOffset() = lineOffset
+
+    override fun compareTo(other: TextPointer) =
```
margarita-nedzelska-sonarsource, line 113: How about adding a few unit tests to increase coverage. On the one hand, it's not mandatory, on the other hand, this will be nice.

Btw, if this method is never used in our codebase, how about throwing `UnsupportedOperationException` instead of this implementation?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/dev/AstPrinter.kt
```diff
@@ -107,3 +105,18 @@ fun textPointerAtOffset(psiDocument: Document, startOffset: Int): TextPointer {
...
+        if (this.line == other.line()) lineOffset.compareTo(other.lineOffset())
+        else line.compareTo(other.line())
+}
+
+private class DefaultTextRange(val start: TextPointer, val end: TextPointer) : TextRange {
+    override fun start() = start
+    override fun end() = end
+    override fun overlap(another: TextRange) = end > another.start() && another.end() > start
```
margarita-nedzelska-sonarsource, line 121: The same as for DefaultTextPointer. compareTo() method

margarita-nedzelska-sonarsource (approved):
Looks good! Just one small suggestion to handle Code Coverage.

And I think commit message should be about migration to the new SQ Api

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/dev/AstPrinter.kt
johann-beleites-sonarsource: You are right and actually I found we can simplify this even more by removing the dependency on the commons `TextRange` & `TextPointer`. There is no real need for it and it removes the methods we wouldn't be using anyway.

PR #65, opened 2021-08-16
SONARKT-42 Fix bug where duplicate strings in annotations are not pro…

johann-beleites-sonarsource:
…perly ignored.


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #67, opened 2021-08-17
SONARKT-81 Add missing ITs

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
Created a [ticket](https://jira.sonarsource.com/browse/SONARKT-114) to move all ITs to Kotlin and Junit 5. Wanted to do it here, but looks like we should not mix it with this current task.

johann-beleites-sonarsource (approved):

PR #71, opened 2021-09-06
SONARKT-68 Fixing S5324 FNs

johann-beleites-sonarsource:
Matching also on super types when calling methods that access Android external storage to avoid FNs when the method called actually belongs to a subtype of Context or Environment


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #74, opened 2021-09-21
SONARKT-142 Rule S5320: Broadcasting intents is security-sensitive

margarita-nedzelska-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/AndroidBroadcastingCheck.kt
```diff
@@ -0,0 +1,95 @@
...
+    override fun visitFunctionCall(
+        callExpression: KtCallExpression,
+        resolvedCall: ResolvedCall<*>,
+        kotlinFileContext: KotlinFileContext
+    ) {
+        val arguments = resolvedCall.valueArgumentsByIndex ?: return
+        val name = resolvedCall.resultingDescriptor.name.asString()
+
+        if (
+            with(kotlinFileContext.bindingContext) {
```
margarita-nedzelska-sonarsource, line 69: I really missed Scala implicits while implementing this check. What do you think about such a move?

sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S5320.html
```diff
@@ -0,0 +1,27 @@
+<p>In Android applications, broadcasting intents is security-sensitive. For example, it has led in the past to the following vulnerability:</p>
```
margarita-nedzelska-sonarsource, line 1: The description with example will be updated once we merge the RSPEC

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ApiExtensions.kt
```diff
@@ -309,3 +313,11 @@ fun Call.findCallInPrecedingCallChain(matcher: FunMatcher, bindingContext: Bindi
...
     return receiver to receiverResolved
 }
+
+fun ResolvedValueArgument.isNull(bindingContext: BindingContext) = (
+    (this as? ExpressionValueArgument)
+        ?.valueArgument
+        ?.getArgumentExpression()
+        ?.predictRuntimeValueExpression(bindingContext)
+        as? KtConstantExpression
+    )?.isNull() ?: false
```
johann-beleites-sonarsource, line 323: `KtExpression` actually has `isNull()` defined on it (as extension method). So you should be able to remove the cast here.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/AndroidBroadcastingCheck.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.jetbrains.kotlin.resolve.calls.model.ResolvedValueArgument
+import org.sonar.check.Rule
+import org.sonarsource.kotlin.api.CallAbstractCheck
+import org.sonarsource.kotlin.api.FunMatcher
+import org.sonarsource.kotlin.api.isNull
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+private const val MESSAGE = "Make sure that broadcasting intents is safe here."
+
+private val MATCHER = FunMatcher(qualifier = "android.content.Context") {
```
johann-beleites-sonarsource, line 34: Really minor, maybe a more descriptive name? E.g. `FUNS_TO_VISIT` or something

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/AndroidBroadcastingCheck.kt
```diff
@@ -0,0 +1,95 @@
...
+            }
+        ) {
+            kotlinFileContext.reportIssue(callExpression.calleeExpression!!, MESSAGE)
+        }
+    }
+
+    private fun BindingContext.isSendOrderedBroadcastAsUser(
+        name: String,
+        argumentsByIndex: List<ResolvedValueArgument>
+    ) = name == "sendOrderedBroadcastAsUser" && (argumentsByIndex.size > 2 && argumentsByIndex[2].isNull(this))
```
johann-beleites-sonarsource, line 84: You could also use `argumentsByIndex.getOrNull(2)` to avoid the explicit range check. Perhaps a matter of taste, do what you think is nicer. If you do decide to change it, obviously same comment for the funs below.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/AndroidBroadcastingCheck.kt
```diff
@@ -0,0 +1,95 @@
...
+        val arguments = resolvedCall.valueArgumentsByIndex ?: return
+        val name = resolvedCall.resultingDescriptor.name.asString()
+
+        if (
+            with(kotlinFileContext.bindingContext) {
+                isSendBroadcast(name, arguments)
+                    || isSendBroadcastAsUser(name, arguments)
+                    || isSendOrderedBroadcast(name, arguments)
+                    || isSendOrderedBroadcastAsUser(name, arguments)
+                    || name in STICKY_BROADCAST_NAMES
```
johann-beleites-sonarsource, line 74: Tiny performance optimization: you could put the check `name in STICKY_BROADCAST_NAMES` first, as the other checks do some data flow analysis.

johann-beleites-sonarsource (approved):
Looks good! A couple of comments here & there, nothing major.

You can also update the metadata already, the rspec PR can be merged.

LGTM!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/AndroidBroadcastingCheck.kt
margarita-nedzelska-sonarsource: I think I can even inline it. Since we have a property it won't be recomputed

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/AndroidBroadcastingCheck.kt
margarita-nedzelska-sonarsource: Yep both solutions are not perfect you either check size or treat null :( But I'll use getOrNull to be consistent with the other checks you've implemented

PR #76, opened 2021-09-22
Use resolved call args instead of call args

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
Should be rebased

PR #79, opened 2021-09-23
SONARKT-139 S1128: FP despite reference to constant in companion obje…

margarita-nedzelska-sonarsource:
…ct of target class


johann-beleites-sonarsource (approved):
Interesting how the compiler imports companion object content! LGTM!

PR #80, opened 2021-09-27
Improve KtChecksVisitor by making the flattening method tail-recursiv…

johann-beleites-sonarsource:
…e and hence safer for large trees


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/visiting/KtChecksVisitor.kt
```diff
@@ -28,7 +28,7 @@ import org.sonarsource.kotlin.plugin.KotlinFileContext
 class KtChecksVisitor(val checks: Checks<out AbstractCheck>) : KotlinFileVisitor() {
 
     override fun visit(kotlinFileContext: KotlinFileContext) {
-        flattenNodes(listOf(kotlinFileContext.ktFile)).let { flatNodes ->
+        flattenNodes(sequenceOf(kotlinFileContext.ktFile)).let { flatNodes ->
             checks.all().forEach { check ->
                 flatNodes.forEach {
```
margarita-nedzelska-sonarsource, line 33: As we are already touching it, let's switch the order. First, iterate over nodes, then over checks?

margarita-nedzelska-sonarsource (approved):
LGTM!

PR #83, opened 2021-09-27
SONARKT-145 Rule S4347: "SecureRandom" seeds should not be predictable 

johann-beleites-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -143,14 +143,15 @@ class FunMatcherBuilderContext(
...
-    var arguments: MutableList<List<ArgumentMatcher>> = mutableListOf(),
+    arguments: List<List<ArgumentMatcher>> = listOf(),
     var definingSupertype: String? = null,
     var matchConstructor: Boolean = false,
     var dynamic: Boolean? = null,
     var extensionFunction: Boolean? = null,
     var suspending: Boolean? = null,
     var returnType: String? = null,
 ) {
+    var arguments: MutableList<List<ArgumentMatcher>> = arguments.toMutableList()
```
margarita-nedzelska-sonarsource, line 154: Good catch to remove "MutableList" from the API! 👍

margarita-nedzelska-sonarsource (approved):
LGTM!

PR #84, opened 2021-09-28
Fix Ruling

margarita-nedzelska-sonarsource:

PR #85, opened 2021-09-29
SONARKT-148 Let the ruling and integration tests fail fast to detect …

johann-beleites-sonarsource:
…errors that occur in the engine.

### DIFF
--- a/sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
+++ b/sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
@@ -261,7 +261,7 @@ internal class KotlinSensorTest : AbstractSensorTest() {
     }
 
     @Test
-    fun `ensure failFast is not triggering when not set`() {
+    fun `ensure failFast is not triggering when set to false`() {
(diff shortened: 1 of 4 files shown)


johann-beleites-sonarsource:
Depends on #84

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
```diff
@@ -208,7 +232,60 @@ internal class KotlinSensorTest : AbstractSensorTest() {
...
+    @Test
+    fun `ensure failFast is not triggering when not set`() {
+        val sensor = failFastSensorWithEnvironmentSetup(false)
+        assertDoesNotThrow { sensor.execute(context) }
+        assertThat(logTester.logs(LoggerLevel.ERROR))
+            .containsExactly("Cannot analyse 'file1.kt' with 'KtChecksVisitor': This is a test message")
+    }
+
+    @Test
+    fun `ensure failFast is not triggered when unset`() {
```
margarita-nedzelska-sonarsource, line 272: Looks like these 2 test names are confused. should these be "not set" -> null and "unset" -> false. Or maybe choose different names to make it less confusing)

margarita-nedzelska-sonarsource (approved):
Other than that looks good!

PR #87, opened 2021-09-30
SONARKT-125 S1143: jump statements should not occur in finally blocks

dorian-burihabwa-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        val visitor = FinallyBlockVisitor(report)
+        finallySection.accept(visitor)
+        visitor.jumpsToLabel
+            .filter { isEscapingTheBlock(it, visitor.labelsInFinallyBlock) }
+            .forEach(report)
+
+        finallySection
+            .collectDescendantsOfType<KtThrowExpression>()
+            .forEach(report)
```
margarita-nedzelska-sonarsource, line 67: You can move all this logic to the `FinallyBlockVisitor `

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,130 @@
...
+private fun buildReportMessage(element: KtElement) =
+    "Remove this %s statement from this finally block.".format(element.toString().lowercase())
+
+@Rule(key = "S1143")
+class ReturnInFinallyCheck : AbstractCheck() {
+
+    override fun visitFinallySection(finallySection: KtFinallySection, kotlinFileContext: KotlinFileContext) {
+        val report: (KtElement) -> Unit = {
+            kotlinFileContext.reportIssue(it, buildReportMessage(it))
+        }
```
margarita-nedzelska-sonarsource, line 57: I think It's ok to inline such lambda. And if you refactor `FinallyBlockVisitor` constructor to expect extension lambda like `FinallyBlockVisitor(private val report: KotlinFileContext.(KtElement) -> Unit)` you will only need to write `val visitor = FinallyBlockVisitor { reportIssue(it, buildReportMessage(it)) }`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Checks if the label in the element points to an unexpected label
+ */
+private fun isEscapingTheBlock(element: KtExpressionWithLabel, expectedLabels: List<String>): Boolean {
+    val label = element.getLabelName() ?: return false
+    return label !in expectedLabels
+}
+
+private fun buildReportMessage(element: KtElement) =
```
margarita-nedzelska-sonarsource, line 48: We prefer moving such private functions at the end of a file.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        override fun visitLabeledExpression(expression: KtLabeledExpression) {
+            expression.name?.let { labelsInFinallyBlock.add(it) }
+            super.visitLabeledExpression(expression)
+        }
+
+        override fun visitLambdaExpression(lambdaExpression: KtLambdaExpression) {
+            lambdaDepthCounter++
+            super.visitLambdaExpression(lambdaExpression)
+            lambdaDepthCounter--
```
margarita-nedzelska-sonarsource, line 84: I think it will be nice to extract a repeated patterns into a separate function, to make code more concise

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
+import org.sonar.check.Rule
+import org.sonarsource.kotlin.api.AbstractCheck
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+/**
+ * Checks if the label in the element points to an unexpected label
+ */
+private fun isEscapingTheBlock(element: KtExpressionWithLabel, expectedLabels: List<String>): Boolean {
+    val label = element.getLabelName() ?: return false
```
margarita-nedzelska-sonarsource, line 44: If this case is impossible, you'll be unable to cover it. So it's ok to use "!!" and add a meaningful comment above

margarita-nedzelska-sonarsource (changes requested):
Looks good!
See my comments, we can have a look at them together.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,119 @@
...
+    private var lambdaDepthCounter = 0
+    private val labelsInFinallyBlock = ArrayList<String>()
+    private val jumpsToLabel = ArrayList<KtExpressionWithLabel>()
+
+    override fun visitFinallySection(finallySection: KtFinallySection) {
+        super.visitFinallySection(finallySection)
+        jumpsToLabel
+            .filter { isEscapingTheBlock(it, labelsInFinallyBlock) }
+            .forEach(report)
+    }
```
margarita-nedzelska-sonarsource, line 48: You can have `try()/finally()` inside the `finally()` block. And this code will report an issue twice.

To my mind, you don't need to wait till the end of finally block, but report directly on labeled expressions, because you cannot escape the label, that is out of your scope. In that case you don't need `jumpsToLabel` property at all

kotlin-checks-test-sources/src/main/kotlin/checks/ReturnInFinallyCheckSample.kt
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+
+        fun returnInLambdaInFinally() {
+            while (true) {
+                try {
+                    throw RuntimeException()
+                } finally {
+                    listOf(1, 2, 3).forEach {
+                        if (it == 0) return // Compliant
```
margarita-nedzelska-sonarsource, line 207: Sorry, I haven't noticed this previously:

this should not be compliant, because `return` without a label will return from the closest named function (described with keyword `fun`) So return in such cases are noncompliant.

And such code should be compliant:

```
listOf(1, 2, 3).forEach (fun f(it: Int) {
                        if (it == 0) return // Compliant
                   })
```

or 

```
listOf(1, 2, 3).forEach {
                        if (it == 0) return@forEach // Compliant
                    }
```

Here you can find more: https://kotlinlang.org/docs/returns.html#return-at-labels

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * Checks if the label in the element points to an unexpected label
+ */
+private fun isEscapingTheBlock(element: KtExpressionWithLabel, expectedLabels: List<String>): Boolean {
+    // Because we only call this function on expressions that have a label, we can assume this is safe
+    val label = element.getLabelName()!!
+    return label !in expectedLabels
+}
```
margarita-nedzelska-sonarsource, line 109: How about?

```suggestion
private fun isEscapingTheBlock(element: KtExpressionWithLabel, expectedLabels: List<String>): Boolean =
    // Because we only call this function on expressions that have a label, we can assume this is safe
    element.getLabelName()!! !in expectedLabels
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ReturnInFinallyCheck.kt
```diff
@@ -0,0 +1,119 @@
...
+        super.visitExpressionWithLabel(expression)
+    }
+}
+
+@Rule(key = "S1143")
+class ReturnInFinallyCheck : AbstractCheck() {
+
+    override fun visitFinallySection(finallySection: KtFinallySection, kotlinFileContext: KotlinFileContext) {
+        val visitor = FinallyBlockVisitor { kotlinFileContext.reportIssue(it, it.buildReportMessage()) }
+        finallySection.accept(visitor)
```
margarita-nedzelska-sonarsource, line 98: How about?

```suggestion
        finallySection.accept(
            FinallyBlockVisitor {
                kotlinFileContext.reportIssue(it, it.buildReportMessage())
            }
        )
```

margarita-nedzelska-sonarsource (changes requested):
A few things to improve and I've noticed some uncovered cases, that I missed previously.

I've seen you haven't applied my suggestions with extension functions, that's not mandatory, but makes code very clean and simple.

PR #92, opened 2021-10-04
SONARKT-120 Rule S2151: "runFinalizersOnExit" should not be called

alban-auzeill:


alban-auzeill:
It's not easy to add one issue in the ruling, it needs the property `sonar.java.jdkhome` to be set to a jdk < 11, otherwise `runFinalizersOnExit` does not exists

johann-beleites-sonarsource (approved):
LGTM!

PR #122, opened 2021-10-29
SONARKT-184 Fix java.lang.IllegalArgumentException in rule S1874 when…

margarita-nedzelska-sonarsource:
… reporting on Enum constructor call


johann-beleites-sonarsource (approved):
LGTM!

PR #123, opened 2021-10-29
SONARKT-172 Fix IllegalArgumentException in FunMatcher, when inferred…

margarita-nedzelska-sonarsource:
… return type is intersection

Still cannot reproduce the issue in full semantic mode and without it. I checked intersection types  and and they seems to be ok.

The issue happens when for some reason the function descriptor is present but it's constructor descriptor is null. IT looks like the problem with the incomplete semantic.

Let's anyway merge the fix, because it was already reported at least 2 times


johann-beleites-sonarsource (approved):
Shame we can't reproduce it. Indeed, let's get this fixed anyway!

PR #124, opened 2021-11-04
SONARKT-185 Prepare Release of SonarKotlin 2.6

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
The new detekt version is RC, but as far as we're using it only to generate rule mappings, I think it's safe to use it

johann-beleites-sonarsource (approved):

PR #125, opened 2021-11-04
SONARKT-182 S1128: FP on imports for delegates and "in" "!in" operators

margarita-nedzelska-sonarsource:
should be rebased


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -141,23 +150,68 @@ class UnnecessaryImportsCheck : AbstractCheck() {
...
+                if (imp.importedName?.asString() != "invoke") true
+                else {
+                    imp.importedFqName !in calls.mapNotNull { it.getResolvedCall(bindingContext)?.resultingDescriptor?.fqNameOrNull() }
+                }
+            }
+        }
 }
 
-private class DataCollector(val file: KtFile) : KtTreeVisitorVoid() {
+private class DataCollector(val file: KtFile, val bindingContext: BindingContext) : KtTreeVisitorVoid() {
```
johann-beleites-sonarsource, line 196: I think you don't need the `bindingContext` here (and correspondingly in the method that instantiates the `DataCollector`)

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -141,23 +150,68 @@ class UnnecessaryImportsCheck : AbstractCheck() {
...
+        }
+
+    private fun getInvokeCallsImportsFilter(calls: Collection<KtCallExpression>, bindingContext: BindingContext) =
+        if (bindingContext == BindingContext.EMPTY) {
+            { imp: KtImportDirective -> imp.importedName?.asString() != "invoke" }
+        } else {
+            { imp: KtImportDirective ->
+                if (imp.importedName?.asString() != "invoke") true
+                else {
+                    imp.importedFqName !in calls.mapNotNull { it.getResolvedCall(bindingContext)?.resultingDescriptor?.fqNameOrNull() }
```
johann-beleites-sonarsource, line 190: Here we execute the `calls.mapNotNull { ... }` for every invocation of the filter, although the result of the `map` is the same every time. WDYT of putting the `calls.mapNotNull { ... }` outside of the lambda in the `else` block, where it can be calculated once and referred to from inside the lambda. The same way it is also done for delegates & arrays above.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -141,23 +150,68 @@ class UnnecessaryImportsCheck : AbstractCheck() {
...
+                            propDescriptor
+                                .accessors
+                                .mapNotNull {
+                                    bindingContext.get(BindingContext.DELEGATED_PROPERTY_RESOLVED_CALL, it)
+                                        ?.resultingDescriptor
+                                        ?.fqNameOrNull()
+                                }
+                        }
+                    } ?: emptyList()
+            }.let { resolvedArrayAccesses ->
```
johann-beleites-sonarsource, line 176: I the name `resolvedArrayAccesses` is a copy+paste artifact :)

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -112,16 +123,14 @@ class UnnecessaryImportsCheck : AbstractCheck() {
         return relevantImports
     }
 
-    private fun collectReferences(file: KtFile) =
+    private fun collectReferences(file: KtFile, bindingContext: BindingContext) =
```
johann-beleites-sonarsource, line 126: Don't need `bindingContext` if `DataCollector` doesn't need it

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -50,30 +56,35 @@ class UnnecessaryImportsCheck : AbstractCheck() {
...
         context: KotlinFileContext
     ) = file.importDirectives.filter { imp ->
         // 1. Filter out & report all imports that import from kotlin.* or the same package as our file
         if (imp.isImportedImplicitlyAlready(file.packageDirective?.name)) {
             imp.importedReference?.let { context.reportIssue(it, MESSAGE_REDUNDANT) }
             false
         } else true
-    }.groupBy {
+    }
+    .groupBy {
```
johann-beleites-sonarsource, line 87: An accidental `\n`?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -187,9 +251,14 @@ private fun KtImportDirective.isImportedImplicitlyAlready(containingPackage: Str
...
 private fun KtReferenceExpression.importableSimpleName() =
     when (this) {
-        is KtOperationReferenceExpression -> operationSignTokenType
-            ?.let { (OperatorConventions.BINARY_OPERATION_NAMES[it] ?: OperatorConventions.UNARY_OPERATION_NAMES[it])?.asString() }
-            ?: getReferencedName()
+        is KtOperationReferenceExpression ->
+            operationSignTokenType?.let { token ->
+                (deparenthesize().parent as? KtPrefixExpression)
+                    ?.let { OperatorConventions.UNARY_OPERATION_NAMES[token] }
+                    ?: (OperatorConventions.getNameForOperationSymbol(token))
```
johann-beleites-sonarsource, line 258: I had a hard time understanding why there is a `?.let` here without using `it` that this is really an if-else block. I suggest using an explicit if-else here:
```suggestion
                if (deparenthesize().parent is KtPrefixExpression) OperatorConventions.UNARY_OPERATION_NAMES[token]
                else OperatorConventions.getNameForOperationSymbol(token)
```

johann-beleites-sonarsource (changes requested):
Nice work, this should greatly improve our imports rule! A couple of suggestions, see comments.

johann-beleites-sonarsource (approved):
LGTM!

PR #127, opened 2021-11-08
SONARKT-186 Fix slow analysis speeds due to re-computation of semantics

margarita-nedzelska-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/converter/KotlinTree.kt
```diff
@@ -30,47 +30,38 @@ import org.sonar.api.batch.fs.InputFile
...
+        psiFile.viewProvider.document ?: throw ParseException("Cannot extract document")
+    } catch (e: AssertionError) {
+        // A KotlinLexerException may occur when attempting to read invalid files
+        throw ParseException("Cannot correctly map AST with a null Document object")
+    }
+    
+    checkParsingErrors(psiFile, document, inputFile)
+    
+    return KotlinSyntaxStructure(psiFile, document, inputFile)
+}
```
johann-beleites-sonarsource, line 50: To me this looks like a factory method - is there a reason why it shouldn't be inside of the `KotlinSyntaxStructure` class / its companion object to associate it with the type it actually produces?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -93,27 +101,52 @@ class KotlinSensor(
         }
         statistics.log()
     }
-
+
```
johann-beleites-sonarsource, line 104: Some unnecessary spaces

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -93,27 +101,52 @@ class KotlinSensor(
...
+    
     private fun analyseFiles(
         sensorContext: SensorContext,
         inputFiles: Iterable<InputFile>,
         progressReport: ProgressReport,
         visitors: List<KotlinFileVisitor>,
         statistics: DurationStatistics,
-    ): Boolean {
+        filenames: List<String>,
+
```
johann-beleites-sonarsource, line 112: You can get rid of this empty line

johann-beleites-sonarsource (approved):
Looks good! I have a few minor suggestions, then LGTM!

PR #138, opened 2021-11-15
Fix quality flaw: Add missing test case for KotlinSensor (exception w…

margarita-nedzelska-sonarsource:
…hen reading kotlin file)


johann-beleites-sonarsource (approved):

johann-beleites-sonarsource:
In principle looks good - just makes the Windows build fail. Do we handle things differently in Windows?

PR #140, opened 2021-11-22
SONARKT-187 Filter out any regular expressions that use string interp…

johann-beleites-sonarsource:
…olation to avoid FPs. This is a temporary workaround until we support string templates in regular expressions.


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #141, opened 2021-11-22
SONARKT-183 Introduce workaround in form of a filter for bad compiler…

johann-beleites-sonarsource:
… diagnostics where the Kotlin compiler erroneously reports unused local variables with partial semantics. The filter simply ignores all unused local variable reports for variables that have the type 'kotlin.Nothing'.


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #148, opened 2021-11-26
SONARKT-182 S1128: Fix FP on unresolved imports

margarita-nedzelska-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -52,22 +53,31 @@ class UnnecessaryImportsCheck : AbstractCheck() {
...
         file: KtFile,
         groupedReferences: Map<String?, List<KtReferenceExpression>>,
         kDocLinks: Collection<String>,
+        unresolvedImports: List<KtImportDirective>,
         arrayAccessesImportsFilter: (KtImportDirective) -> Boolean,
         context: KotlinFileContext
-    ) = file.importDirectives.filter { imp ->
+    ) = file.importDirectives
+        // 0. Filter out unresolved imports, to avoid FPs in case of incomplete semantic
+        .filter { it !in unresolvedImports }
```
johann-beleites-sonarsource, line 79: That's a nice trick to avoid FPs!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -52,22 +53,31 @@ class UnnecessaryImportsCheck : AbstractCheck() {
...
+        .filter { imp ->
         // 1. Filter out & report all imports that import from kotlin.* or the same package as our file
         if (imp.isImportedImplicitlyAlready(file.packageDirective?.name)) {
```
johann-beleites-sonarsource, line 82: The indentation is off here, due to a mix of using 
```
... }
.foo { bar }
```
and
```
...
}.foo {
    bar
}
```
styles. I think we should stay consistent, then the indentation can also stay consistent.

johann-beleites-sonarsource (approved):
A small styling issue, otherwise LGTM!

PR #150, opened 2021-12-02
SONARKT-192 Pass down the file name to the Kotlin compiler, s.t. erro…

johann-beleites-sonarsource:
…r messages become more meaningful


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/converter/KotlinTree.kt
```diff
@@ -36,7 +36,8 @@ data class KotlinSyntaxStructure(val ktFile: KtFile, val document: Document, val
     companion object {
         @JvmStatic
         fun of(content: String, environment: Environment, inputFile: InputFile): KotlinSyntaxStructure {
-            val psiFile: KtFile = environment.ktPsiFactory.createFile(normalizeEol(content))
+
+            val psiFile: KtFile = environment.ktPsiFactory.createFile(inputFile.filename(), normalizeEol(content))
```
margarita-nedzelska-sonarsource, line 40: Looks ok, I am just not sure if we want to have just a filename or a path to the file?

johann-beleites-sonarsource: Yea of course, if in doubt that can be a lot more helpful

margarita-nedzelska-sonarsource (approved):

PR #153, opened 2021-12-07
Update dependencies of utils-kotlin module

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
Looks like an old gson dependency is coming from utils-kotlin.
And red whitesource scan doesn't allow me to do a release :(

johann-beleites-sonarsource (approved):

PR #157, opened 2021-12-15
SONARKT-195 Catching any exceptions that may occur during BindingCont…

johann-beleites-sonarsource:
…ext generation.


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #169, opened 2022-03-14
Use getStringArray instead of get for the Java libraries and binaries

johann-beleites-sonarsource:


m-g-sonar (approved):
The change LGTM!

If it fixes an issue, I would create a ticket, to include it in our release note and track the change.

PR #172, opened 2022-07-06
Fix issue with compiler failing when multiple dots appear in file descriptor

chrislain-razafimahefa-sonarsource:


johann-beleites-sonarsource (approved):
Looks good, LGTM! I've created a ticket to track this issue: SONARKT-211

PR #174, opened 2022-07-07
SONARKT-197 S6316 should not crash when job declaration is not directly followed by a call to delay.

chrislain-razafimahefa-sonarsource:


johann-beleites-sonarsource (approved):
Nice, LGTM!

PR #175, opened 2022-07-11
SONARKT-199 S4830 misses exceptions being thrown in catch clauses

chrislain-razafimahefa-sonarsource:


chrislain-razafimahefa-sonarsource:
For now, I have not been able to improve the code coverage.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -19,44 +19,107 @@
...
+        private var throwFound: Boolean = false
+        private var catchFound: Boolean = false
+
+        override fun visitThrowExpression(expression: KtThrowExpression) {
+            val callExpr = expression.thrownExpression as? KtCallExpression
+            throwFound = callExpr?.let {
+                ConstructorMatcher {
+                    qualifier = "java.security.cert.CertificateException"
+                    withNoArguments()
+                }.matches(it, bindingContext)
```
johann-beleites-sonarsource, line 106: We should be able to make our lives a lot easier with a couple of little helpers, now and in future, and get more accurate results at the same time. Copy & paste these two helpers into `ApiExtensions.kt`:
```
fun PsiElement?.getType(bindingContext: BindingContext): KotlinType? =
    this?.let {
        when(this) {
            is Call -> getResolvedCall(bindingContext)?.getReturnType()
            is KtCallExpression -> getResolvedCall(bindingContext)?.getReturnType()
            is KtParameter -> determineType(bindingContext)
            is KtTypeReference -> determineType(bindingContext)
            is KtProperty -> determineType(bindingContext)
            is KtDotQualifiedExpression -> resolveReferenceTarget(bindingContext).getType()
            is KtReferenceExpression -> bindingContext.get(BindingContext.REFERENCE_TARGET, this).getType()
            else -> bindingContext.get(BindingContext.VARIABLE, it)?.type
        }

    }

fun DeclarationDescriptor?.getType(): KotlinType? =
    this?.let {
        when(this) {
            is FunctionDescriptor -> returnType
            is ClassDescriptor -> classValueType
            else -> null
        }
    }
```
Then, you can simply replace the body of this function with:
```
            if (
                expression.thrownExpression.getType(bindingContext)
                    ?.getJetTypeFqName(false) == "java.security.cert.CertificateException"
            ) {
                throwFound = true
            }
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -19,44 +19,109 @@
...
+            } ?: false
+        }
+
+        override fun visitCatchSection(catchClause: KtCatchClause) {
+            catchFound = CERTIFICATE_EXCEPTION == catchClause.catchParameter?.typeReference?.typeElement?.text
```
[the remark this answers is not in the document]
johann-beleites-sonarsource: Instead, what I can recommend is use the same helpers we used above (and maybe extract the string into a field, as you've already done with the short string):
```
            catchFound = catchClause.catchParameter.getType(bindingContext)
                ?.getJetTypeFqName(false) == "java.security.cert.CertificateException"
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -19,44 +19,107 @@
...
+
+    /*
+     * Returns true if a function contains a call to "checkClientTrusted" or "checkServerTrusted".
+     */
+    private fun KtNamedFunction.callsCheckTrusted(bindingContext: BindingContext): Boolean {
+        val visitor = object : KtVisitorVoid() {
+            private var foundCheckTrustedCall: Boolean = false
+
+            override fun visitCallExpression(expression: KtCallExpression) {
+                foundCheckTrustedCall = funMatchers.any { it.matches(expression, bindingContext) }
```
johann-beleites-sonarsource, line 76: Be careful, you still need a conditional statement here. Add this in the Sample file after line 90 to test this case:
```
                getAcceptedIssuers()
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -19,44 +19,107 @@
...
+            && !function.throwsCertificateExceptionWithoutCatching(bindingContext)
+        ) {
+            kotlinFileContext.reportIssue(function.nameIdentifier ?: function,
+                "Enable server certificate validation on this SSL/TLS connection.")
+        }
+    }
+
+    private fun KtNamedFunction.belongsToTrustManagerClass(bindingContext: BindingContext): Boolean {
+        return funMatchers.any { it.matches(this, bindingContext) }
+    }
```
johann-beleites-sonarsource, line 66: You can simplify functions like this to expression functions:
```
    private fun KtNamedFunction.belongsToTrustManagerClass(bindingContext: BindingContext) =
        funMatchers.any { it.matches(this, bindingContext) }
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -19,44 +19,107 @@
...
+        val funMatchers = listOf(
+            FunMatcher {
+                definingSupertype = "javax.net.ssl.X509TrustManager"
+                withNames("checkClientTrusted", "checkServerTrusted")
+            },
+            FunMatcher {
+                definingSupertype = "javax.net.ssl.X509ExtendedTrustManager"
+                withNames("checkClientTrusted", "checkServerTrusted")
+            })
     }
```
johann-beleites-sonarsource, line 50: Another artifact from this being an older rule: We have actually stopped declaring 'static' values such as this one inside of the companion object. Instead, we usually declare them above the class, where you also correctly declared the const for `CertificateException`. Since we are touching this, we might as well get rid of the companion object here.

johann-beleites-sonarsource (changes requested):
A couple of suggestions, see the comments below. One major improvement we could make to the codebase now and in future is introduce a couple of helpers to help with a problem you are facing in this rule as well. I've added a comment about it, with the suggested helper. Feel free to let me know if you want to talk about any points!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
chrislain-razafimahefa-sonarsource: These snippets are and will be very useful. Thanks for sharing.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
johann-beleites-sonarsource: I think part of the reason was also the fact that companion objects are not really very natural to the language and overall just a bit clunky. Let us discuss this elsewhere in more detail, though, if it interests you!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ApiExtensions.kt
```diff
@@ -453,6 +453,7 @@ fun KtExpression?.getterMatches(bindingContext: BindingContext, propertyName: St
 fun KtBinaryExpression.isPlus() =
     this.operationReference.operationSignTokenType?.let { OperatorConventions.BINARY_OPERATION_NAMES[it] }?.asString() == "plus"
 
+
```
johann-beleites-sonarsource, line 456: Superfluous empty line
```suggestion
```

johann-beleites-sonarsource (approved):
Nice, this rule now looks a lot more understandable than before! LGTM!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ApiExtensions.kt
chrislain-razafimahefa-sonarsource: Fixed.

PR #177, opened 2022-07-14
SONARKT-15 Introduce Kotlin source version property

johann-beleites-sonarsource:
Introduce property that users can set to specify the Kotlin source code version they would like to scan their code with.


chrislain-razafimahefa-sonarsource (approved):
Looks very good. Well done!

PR #182, opened 2022-07-18
SONARKT-198 S1128 FP when imported package is postfix of current package

johann-beleites-sonarsource:


chrislain-razafimahefa-sonarsource (approved):
Good catch! LGTM!

PR #185, opened 2022-07-19
Cleaning up constants that were spread across multiple files

johann-beleites-sonarsource:


chrislain-razafimahefa-sonarsource (approved):
Sounds good!

PR #186, opened 2022-07-19
SONARKT-202 S1871 should not consider two method calls the same if they're calling different methods with the same name

chrislain-razafimahefa-sonarsource:
- Fix the ticket issue


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/api/ApiExtensionsKtTest.kt
```diff
@@ -260,6 +261,41 @@ class ApiExtensionsKtDetermineTypeTest {
...
+        )
+        bindingContext = kotlinTree.bindingContext
+        ktFile = kotlinTree.psiFile
+    }
+
+    @Test
+    fun `determineSignature of KtQualifiedExpression`() {
+        val expr = ktFile.findDescendantOfType<KtQualifiedExpression> { it.text == "this.prop" }!!
+
+       assertThat(expr.determineSignature(bindingContext)).isNotNull
```
johann-beleites-sonarsource, line 290:
```suggestion
        assertThat(expr.determineSignature(bindingContext)?.fqNameOrNull()?.asString()).isEqualTo("bar.Foo.prop")
```

johann-beleites-sonarsource (approved):
Looks great! Just one little suggestion to improve testing, and then LGTM!

PR #203, opened 2022-10-26
SONARKT-229 Move to versionCatalogs instead of dependency management block

johann-beleites-sonarsource:
- Upgrade to Gradle 7.5.1
- SONARKT-229 Migrate to versionCatalogs for dependency management with Gradle


chrislain-razafimahefa-sonarsource (changes requested):
Well done! It gives more structure to the code. I have no changes to suggest. There is just that the build does not pass on CI.

chrislain-razafimahefa-sonarsource:
Actually, the build is ok now.

chrislain-razafimahefa-sonarsource (approved):

PR #206, opened 2022-10-28
SONARKT-36 Improve detection of 'this' inside of 'with', 'apply' and 'run' scope functions

johann-beleites-sonarsource:


johann-beleites-sonarsource:
Note that the low coverage is due to the chained null-safe calls, which are infeasible to cover without mocking.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/verifier/KotlinVerifier.kt
```diff
@@ -48,7 +48,7 @@ private val DEFAULT_TEST_JARS_DIRECTORY = "../kotlin-checks-test-sources/build/t
 class KotlinVerifier(private val check: AbstractCheck) {
 
     var fileName: String = ""
-    var classpath: List<String> = System.getProperty("java.class.path").split(":") + DEFAULT_KOTLIN_CLASSPATH
+    var classpath: List<String> = System.getProperty("java.class.path").split(System.getProperty("path.separator")) + DEFAULT_KOTLIN_CLASSPATH
```
dorian-burihabwa-sonarsource, line 51: Is this for portability? Or does it help this specific PR? Anyway, it is probably a good change

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 (BTW I am surprised such a change does not affect the ruling)

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/verifier/KotlinVerifier.kt
johann-beleites-sonarsource: Yea, it is for portability (on Windows, splitting by `:` is actually completely breaking the classpath - you need to use `;` instead). To be honest, I am surprised our unit tests all passed on Windows so far :thinking:

PR #207, opened 2022-10-28
SONARKT-228 Remove dependency on apache commons-text in kotlin-utils

margarita-nedzelska-sonarsource:


utils-kotlin/src/test/kotlin/org/sonarsource/kotlin/externalreport/common/TranslatorTest.kt
```diff
@@ -0,0 +1,53 @@
+package org.sonarsource.kotlin.externalreport.common
```
johann-beleites-sonarsource, line 1: Here as well, we should put a license header. I realize that we haven't done so for the other tests in this module, something to update at some point too.

utils-kotlin/src/main/kotlin/org/sonarsource/kotlin/externalreport/common/Translator.kt
```diff
@@ -0,0 +1,394 @@
...
+                    if (Character.isLowSurrogate(c2)) {
+                        out.write(c2.code)
+                        pos++
+                    }
+                }
+                continue
+            }
+            // contract with translators is that they have to understand codepoints
+            // and they just took care of a surrogate pair
+            for (pt in 0 until consumed) {
```
johann-beleites-sonarsource, line 65: I was a little confused here at first, I think it could be easier to understand if we put the `for` loop into an `else` block and remove the `continue`:
```suggestion
            } else {
                // contract with translators is that they have to understand codepoints
                // and they just took care of a surrogate pair
                for (pt in 0 until consumed) {
```
(The suggestion is incomplete)

utils-kotlin/src/main/kotlin/org/sonarsource/kotlin/externalreport/common/Translator.kt
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    fun translate(input: CharSequence, out: Writer) {
+        var pos = 0
+        val len = input.length
+        while (pos < len) {
+            val consumed = translate(input, pos, out)
+            if (consumed == 0) {
+                // inlined implementation of Character.toChars(Character.codePointAt(input, pos))
+                // avoids allocating temp char arrays and duplicate checks
```
johann-beleites-sonarsource, line 50: I don't quite understand this comment. Why is the `Character.toChars(Character.codePointAt(input, pos))` relevant here?

johann-beleites-sonarsource (changes requested):
Nice job on eliminating the dependency! Just a couple of suggestions

utils-kotlin/src/main/kotlin/org/sonarsource/kotlin/externalreport/common/Translator.kt
```diff
@@ -0,0 +1,394 @@
+package org.sonarsource.kotlin.externalreport.common
```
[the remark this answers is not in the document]
margarita-nedzelska-sonarsource: The algorithm itself is coming from Apache Commons. But I simplified it a little bit. I was about to ask if we're subject to an attribution here.

Do you know what is expected in this case?

utils-kotlin/src/main/kotlin/org/sonarsource/kotlin/externalreport/common/Translator.kt
```diff
@@ -0,0 +1,289 @@
...
+object Translator {
+
+    fun escapeHtml4(input: CharSequence) =
+        StringWriter(input.length * 2).apply {
+            (input.indices).forEach { i ->
+                TO_ESCAPE[input[i]]?.let {
+                    write(it)
+                } ?: write(input[i].toString())
+            }
+        }.toString()
```
johann-beleites-sonarsource, line 33: This is a lot more readable than the library implementation, nice!
I have a couple of different suggestions on how we could simplify it even one more step. Since this is not in production and we are only parsing relatively small amounts of data, it doesn't need to be ultra efficient IMO. Considering that, here's my first suggestion:
```suggestion
        input.map { TO_ESCAPE[it] ?: it }.joinToString("")
```
Although I think it's not necessary, if we want to keep the `StringWriter`, we could still simplify it a little:
```suggestion
        StringWriter(input.length * 2).apply {
            input.forEach { char ->
                write(TO_ESCAPE[char] ?: char.toString())
            }
        }.toString()
```
```

johann-beleites-sonarsource (commented):
Looks good, and very nice to get rid of a whole library for the little code we need instead. I have a couple of proposals of how I think we can simplify it even further.

johann-beleites-sonarsource (approved):
Nice, LGTM!

PR #208, opened 2022-10-28
SONARKT-209 excluded local IPv4-mapped IPv6 addresses for rule S1313

leonardo-pilastri-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/HardcodedIpCheck.kt
```diff
@@ -85,7 +87,13 @@ private fun isValidIPV6(ipv6: String, ipv4: String?): Boolean {
...
 
-private fun isIPV6Exception(ip: String) = IPV6_LOOPBACK.matches(ip) || IPV6_NON_ROUTABLE.matches(ip)
+private fun isIPV6Exception(ip: String) = IPV6_LOOPBACK.matches(ip) || IPV6_NON_ROUTABLE.matches(ip) || isReservedDocumentationIp(ip)
+        || isLocalIPv4MappedIPv6(ip)
 
 private fun getCompressionSeparatorCount(str: String) = str.split("::").size - 1
+
+private fun isReservedDocumentationIp(str: String) : Boolean = RESERVED_DOC_IP_RANGE.any { it -> str.startsWith(it) }
+
+private fun isLocalIPv4MappedIPv6(str: String) : Boolean = LOCAL_IPV4_MAPPED_IPV6.any { it -> str.startsWith(it, true) }
```
johann-beleites-sonarsource, line 99: Just a couple of formatting issues:
* There is no space in front of the `:` for declaring types
* Missing newline at the end of the file
If you use the reformat code function of IntelliJ, it should fix those issues automatically for you (`Code -> Reformat Code`)
By reformatting this, we'll stay aligned to the official Kotlin code conventions.

johann-beleites-sonarsource, line 99: If you use `it` as parameter in anonymous functions such as `any`, you don't need to declare the parameter. `it` is the default name and you only need to declare the parameter explicitly, if you want to use a different name or do something more complex with the argument (e.g. destructure it). So for line 97, it would simply become
```
.any { str.startsWith(it) }
```
To make it more readable, you could use a more impactful name for `str`, such as `ip`. (I realize that the existing function on line 95 is not a good example here :))

johann-beleites-sonarsource (changes requested):
Nice work, the logic looks good! There are just a couple of small formatting suggestions to ensure we stay aligned with coding conventions.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/HardcodedIpCheck.kt
```diff
@@ -85,10 +86,14 @@ private fun isValidIPV6(ipv6: String, ipv4: String?): Boolean {
...
 private fun isIPV6Exception(ip: String) = IPV6_LOOPBACK.matches(ip) || IPV6_NON_ROUTABLE.matches(ip) || isReservedDocumentationIp(ip)
+        || isLocalIPv4MappedIPv6(ip)
 
 private fun getCompressionSeparatorCount(str: String) = str.split("::").size - 1
 
-private fun isReservedDocumentationIp(str: String) : Boolean = RESERVED_DOC_IP_RANGE.any { it -> str.startsWith(it) }
\ No newline at end of file
+private fun isReservedDocumentationIp(str: String) : Boolean = RESERVED_DOC_IP_RANGE.any { it -> str.startsWith(it) }
+
+private fun isLocalIPv4MappedIPv6(str: String) : Boolean = LOCAL_IPV4_MAPPED_IPV6.any { it -> str.startsWith(it, true) }
```
johann-beleites-sonarsource, line 99: When you write functions that return a single expression, like here, you can actually omit the return type declaration entirely. So 
```
private fun isReservedDocumentationIp(str: String) : Boolean = RESERVED_DOC_IP_RANGE.any { it -> str.startsWith(it) }
```
would become 
```
private fun isReservedDocumentationIp(str: String) = RESERVED_DOC_IP_RANGE.any { it -> str.startsWith(it) }
```
This makes refactoring much simpler, as types aren't declared redundantly. Note that if you haven't done so already, it helps to enable type hints in the IDE and you'll still be able to see the type while editing the code.

johann-beleites-sonarsource (approved):
LGTM!

PR #210, opened 2022-11-01
SONARKT-117 Add rule S6203

leonardo-pilastri-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/TextBlocksInComplexExpressionsCheck.kt
```diff
@@ -0,0 +1,63 @@
...
+    companion object {
+        const val DEFAULT_LINES_NUMBER = 5
+    }
+
+    @RuleProperty(
+        key = "MaximumNumberOfLines",
+        description = "The maximum number of lines in a text block that can be nested into a complex expression.",
+        defaultValue = "" + DEFAULT_LINES_NUMBER
+    )
+    var linesNumber = DEFAULT_LINES_NUMBER
```
dorian-burihabwa-sonarsource, line 45: Because this rule has a property, we should probably add a test where the value is not the default one

dorian-burihabwa-sonarsource (changes requested):
Nice work. There are some small improvements that would help us with coverage and testing the property.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/TextBlocksInComplexExpressionsCheckTest.kt
```diff
@@ -0,0 +1,34 @@
...
+internal class TextBlocksInComplexExpressionsCheckTest : CheckTest(TextBlocksInComplexExpressionsCheck())
+
+@Test
+fun test() {
+    val check = TextBlocksInComplexExpressionsCheck()
+    check.linesNumber = 1
+    KotlinVerifier(check) {
+        fileName = "TextBlocksInComplexExpressionsCheckSampleSingleLines.kt"
+    }.verify()
+}
```
dorian-burihabwa-sonarsource, line 34: I think it would be best to put the test within the test class so that it is displayed in context in the test results.
We can also rename the test to be more explicit in the property to verify
```suggestion
internal class TextBlocksInComplexExpressionsCheckTest : CheckTest(TextBlocksInComplexExpressionsCheck()) {
  @Test
  fun `custom linesNumber rule property is effectively enforced`() {
      val check = TextBlocksInComplexExpressionsCheck()
      check.linesNumber = 1
      KotlinVerifier(check) {
          fileName = "TextBlocksInComplexExpressionsCheckSampleSingleLines.kt"
      }.verify()
  }
}
```

dorian-burihabwa-sonarsource (changes requested):
Almost there! Let's take some time to structure and document our tests

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #211, opened 2022-11-02
SONARKT-226 Re-enable S125

johann-beleites-sonarsource:
The old approach to detecting commented-out code was trying to parse comments using the Kotlin compiler and if
successful raising an issue. This approach was quite slow and resulted in S125 being disabled silently for a
while. To re-enable this rule, this new implementation makes use of the sonar-analyzer-commons-recognizers
library instead, which only does basic pattern matching but is significantly more performant.


kotlin-checks-test-sources/src/main/kotlin/checks/CommentedCodeCheckSample.kt
```diff
@@ -23,4 +30,22 @@ class CommentedCodeCheckSample {
...
+//            println("Completed: ${valid.size} valid, ${invalid.size} invalid of ${conns.size} total [attempts $attempts]")
+
+    // Noncompliant@+2
+
+//    println("This")
+//    println("is")
+//    println("commented-out")
+//    println("code")
+//    println("and")
+//    println("detected")
```
dorian-burihabwa-sonarsource, line 50: Do we have an example of non-code-parsed-as-code in this sample? If so, it would be nice to point it out (as a comment 🙃 )

its/ruling/src/test/resources/expected/kotlin/kotlin/kotlin-S125.json
```diff
@@ -0,0 +1,208 @@
...
+374,
+],
+'kotlin-kotlin-project:sources/kotlin/kotlin/compiler/backend/src/org/jetbrains/kotlin/codegen/range/forLoop/IteratorForLoopGenerator.kt':[
+67,
+],
+'kotlin-kotlin-project:sources/kotlin/kotlin/compiler/container/tests/org/jetbrains/kotlin/container/tests/ComponentContainerTest.kt':[
+190,
+],
+'kotlin-kotlin-project:sources/kotlin/kotlin/compiler/frontend/src/org/jetbrains/kotlin/cfg/pseudocode/ControlFlowInstructionsGenerator.kt':[
+313,
```
dorian-burihabwa-sonarsource, line 23: This one stumbles starts a line earlier than expected but it does fall on a TP

its/ruling/src/test/resources/expected/kotlin/kotlin/kotlin-S125.json
```diff
@@ -0,0 +1,208 @@
...
+],
+'kotlin-kotlin-project:sources/kotlin/kotlin/compiler/frontend/src/org/jetbrains/kotlin/resolve/calls/CandidateResolver.kt':[
+286,
```
dorian-burihabwa-sonarsource, line 26: Close enough

kotlin-checks-test-sources/src/main/kotlin/checks/CommentedCodeCheckSample.kt
dorian-burihabwa-sonarsource: Do we have any example where the code starts immediately after the inline comment tokens like
```kotlin
//callToMethod()
```
It looks like we could detect [such things](https://github.com/SonarSource/slang-test-sources/blob/ff75c5df9b98cd60850e5622970d195a85709a0f/kotlin/kotlin/idea/idea-core/src/org/jetbrains/kotlin/idea/j2k/IdeaDocCommentConverter.kt#L201) in the previous version of the rule.

its/ruling/src/test/resources/expected/kotlin/intellij-rust/kotlin-S125.json
```diff
@@ -0,0 +1,8 @@
+{
+'kotlin-intellij-rust-project:sources/kotlin/intellij-rust/src/main/kotlin/org/rust/ide/intentions/IfLetToMatchIntention.kt':[
+51,
+],
+'kotlin-intellij-rust-project:sources/kotlin/intellij-rust/src/main/kotlin/org/rust/ide/template/RsContextType.kt':[
```
dorian-burihabwa-sonarsource, line 5: Not really an FP but weird that it does not pick up the 2 previous examples as well

dorian-burihabwa-sonarsource (changes requested):
Good job. It looks like it manages not to raise FPs on the ruling and the implementation is pretty easy to understand.
Once, we get rid of this non-existant documentation tag and remove the unnecessary imports, we should be good to merge 👍🏿

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/converter/KotlinCodeVerifier.kt
```diff
@@ -38,78 +37,128 @@ import org.jetbrains.kotlin.psi.KtPrefixExpression
...
+        "Double",
+        "interface",
+        "Long",
+        "Int",
+        "Float",
+        "super",
+        "true",
+        "false",
+        "object",
+        "companion"
```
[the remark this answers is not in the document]
johann-beleites-sonarsource: No, exhaustiveness is not a goal. I took what we do in Java, removed a couple not relevant to Kotlin, and added others that are common in Kotlin. No doubt it could be improved - I decided to stop optimizing since I've spent a lot of time on it already and it's a never-ending task.

kotlin-checks-test-sources/src/main/kotlin/checks/CommentedCodeCheckSample.kt
dorian-burihabwa-sonarsource: Actually we do have such examples in the ruling it might be related to the expression being short/simple

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/converter/KotlinCodeVerifier.kt
```diff
@@ -38,78 +37,128 @@ import org.jetbrains.kotlin.psi.KtPrefixExpression
...
+    )
+
+    private val COMMON_KOTLIN_PATTERN_SNIPPETS = arrayOf(
+        "for(", "if(", "while(", "catch(", "when(", "try{", "else{", ".let{", ".also{", ".run{", ".apply{", ".map{", ".forEach{", "()"
```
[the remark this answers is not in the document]
johann-beleites-sonarsource: In principle I agree with you, unfortunately, it is currently required by analyzer commons and we simply accept this contract in our other analyzers. I don't think its priority is high enough to warrant changing analyzer commons at the moment.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CommentedCodeCheck.kt
```diff
@@ -54,12 +51,13 @@ class CommentedCodeCheck : AbstractCheck() {
             }
         })
 
-        val codeVerifier = KotlinCodeVerifier()
         groupedComments.forEach { comments ->
-            val content = comments.joinToString("\n") { it.getContent() }
-            if (codeVerifier.containsCode(content)) {
-                val textRanges = comments.map { kotlinFileContext.textRange(it) }
-                kotlinFileContext.reportIssue(kotlinFileContext.merge(textRanges), "Remove this commented out code.")
+            if (file.firstChild !in comments) {
```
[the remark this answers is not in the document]
johann-beleites-sonarsource: The problem is that we can have multiple AST nodes belonging to the header, so when the header comment uses multiple lines of single-line comments (`//`) instead of simply a multi-line comment (`/*...*/`). Unfortunately, this seems to be a thing in the wild. So before grouping them, it's not easy to determine whether a single-line comment that is not the first node in the file, is part of the header.

its/ruling/src/test/resources/expected/kotlin/kotlin/kotlin-S125.json
johann-beleites-sonarsource: Yea, can't get them all :see_no_evil:

its/ruling/src/test/resources/expected/kotlin/intellij-rust/kotlin-S125.json
johann-beleites-sonarsource: It is because it doesn't meet the minimum threshold to satisfy `words > 2`. We've got to make the cutoff somewhere, admittedly a bit unfortunate example, though :shrug:

dorian-burihabwa-sonarsource (approved):
Thanks for the changes. LGTM 👍🏿

PR #212, opened 2022-11-02
SONARKT-233 Support suppress annotation on when-case expressions

dorian-burihabwa-sonarsource:


its/plugin/projects/suppress-warnings/kotlin/file2.kt
```diff
@@ -35,3 +35,79 @@ fun complexity() {
...
+        27 -> "0"
+        28 -> "0"
+        29 -> "0"
+        30 -> "0"
+        else -> "Unknown"
+    }
+}
+
+@Suppress("kotlin:S1479")
+fun matchingValue(value: Int): String {
```
johann-beleites-sonarsource, line 78: Are we sure this parses correctly (the signatures of the two methods seem to be identical)? 
Can we add 1 case where we don't suppress to this file and check we do trigger the rule at least once?

kotlin-checks-test-sources/src/main/kotlin/sample/IssueNonSuppressionSample.kt
```diff
@@ -31,3 +31,43 @@ class Foo2 {
...
+            25 -> "0"
+            26 -> "0"
+            27 -> "0"
+            28 -> "0"
+            29 -> "0"
+            30 -> "0"
+            else -> "Unknown"
+        }
+    }
+}
```
johann-beleites-sonarsource, line 73: Minor: missing newline at end of file

kotlin-checks-test-sources/src/main/kotlin/sample/IssueWithoutSuppressionSample.kt
```diff
@@ -31,3 +31,42 @@ class Foo3 {
...
+            25 -> "0"
+            26 -> "0"
+            27 -> "0"
+            28 -> "0"
+            29 -> "0"
+            30 -> "0"
+            else -> "Unknown"
+        }
+    }
+}
```
johann-beleites-sonarsource, line 72: Minor: missing newline

johann-beleites-sonarsource (commented):
A couple of minor suggestions, after that LGTM!

PR #214, opened 2022-11-03
SONARKT-235 Improve spotless configuration

margarita-nedzelska-sonarsource:
* Update spotless to the latest version 6.11.0
* configure spotless for all modules except kotlin-checks-test-sources

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,4 @@
-import io.spring.gradle.dependencymanagement.dsl.DependencyManagementExtension
+
 import org.sonarsource.kotlin.buildsrc.tasks.CreateRuleStubsTask
 import org.sonarsource.kotlin.buildsrc.tasks.FetchRuleMetadata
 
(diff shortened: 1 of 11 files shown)


margarita-nedzelska-sonarsource:
I haven't changed what was configured before. Just applied it to the whole project.

build.gradle.kts
```diff
@@ -9,13 +9,53 @@ plugins {
...
+        lineEndings = com.diffplug.spotless.LineEnding.UNIX
+
+        fun SourceSet.findSourceFilesToTarget() = allJava.srcDirs.flatMap { srcDir ->
+            project.fileTree(srcDir).filter { file ->
+                file.name.endsWith(".kt") || (file.name.endsWith(".java") && file.name != "package-info.java")
+            }
+        }
+
+        kotlin {
+            // ktlint()
```
dorian-burihabwa-sonarsource, line 34: Can we remove this or add a comment explaining why there might have been a good reason to call `ktlint`?

dorian-burihabwa-sonarsource:
```suggestion
```

sonar-kotlin-plugin/build.gradle.kts
```diff
@@ -80,8 +44,12 @@ test.dependsOn(project(":kotlin-checks-test-sources").tasks.named("build"))
...
 tasks.jar {
     manifest {
-        val displayVersion = if (project.property("buildNumber") == null) project.version else project.version.toString()
-            .substring(0, project.version.toString().lastIndexOf(".")) + " (build ${project.property("buildNumber")})"
+        val displayVersion = if (project.property("buildNumber") == null) {
+            project.version
+        } else {
+            project.version.toString()
+                .substring(0, project.version.toString().lastIndexOf(".")) + " (build ${project.property("buildNumber")})"
+        }
```
dorian-burihabwa-sonarsource, line 52: Thanks, that's a lot easier to read 👍🏿

dorian-burihabwa-sonarsource (approved):
LGTM! If possible, it would be nice to remove the commented ktlint line or add some more information about why we should keep it here. Do we need to run it for every release? every year? whenever we add a new module?

build.gradle.kts
margarita-nedzelska-sonarsource: I think we wanted to use ktlint at some point but then decided not to, as we were not happy with the result.

I think it's okay to drop the line and re-launch the discussion in the future if we want to use it or not.

johann-beleites-sonarsource (commented):
Nice!

PR #217, opened 2022-11-07
[SONARKT-129] Add rule S1206

leonardo-pilastri-sonarsource:


sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1206.html
```diff
@@ -0,0 +1,40 @@
+<p>In Kotlin you must override either both or neither of the <code>equals(Any?)</code> and <code>hashCode()</code> methods in order to keep the
```
johann-beleites-sonarsource, line 1: I found a couple of mistakes in the rule description - I updated the RSPEC PR, you should be able to simply update from there.

johann-beleites-sonarsource (changes requested):
Nice approach, a few (mostly Kotlin-focused) suggestions.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsOverridenWithHashCodeCheck.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun visitClassBody(klass: KtClassBody, ctx: KotlinFileContext) {
+        var equalsMethod: KtNamedFunction? = null
+        var hashCodeMethod: KtNamedFunction? = null
+
+        klass.functions.forEach {
+            when {
+                hashCodeMethod == null && hashCodeMatcher.matches(it, ctx.bindingContext) -> hashCodeMethod = it
+                equalsMethod == null && equalsMatcher.matches(it, ctx.bindingContext) -> equalsMethod = it
+            }
+            if( hashCodeMethod != null && equalsMethod != null ) return@forEach
```
johann-beleites-sonarsource, line 56: This does not do what you intend to do here. `return@forEach` will return from the current lambda, nested in the forEach. In other words, it will return from evaluating the current list item, to then move on to the next. Simply change this to `return` and it will exit the entire `visitCassBody` function, which is what we want to do here.

This is actually an excellent candidate for a new rule!

johann-beleites-sonarsource, line 56: Formatting (spaces after if and not after `(` / before `)`):
```
if (hashCodeMethod != null && equalsMethod != null)
```

johann-beleites-sonarsource, line 56: How come you are using a when and an if construct here? It will be more readable if we use one or the other. By moving the `return` case to be the first `when` case or `if` condition, you avoid one unnecessary check of a function.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsOverridenWithHashCodeCheck.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        if (equalsMethod != null && hashCodeMethod == null) {
```
johann-beleites-sonarsource, line 59: Since we are returning from the method if both are non-null, `(equalsMethod != null) => (hashCodeMethod == null)`. Same is true the other way around, so if `hashCodeMethod != null`, then it is guaranteed that `equalsMethod == null`. So you can remove the second part from the conditions, as it can never be false.

Note that this only holds once we actually return from the method above and SonarLint will flag these conditions to do exactly that.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsOverridenWithHashCodeCheck.kt
```diff
@@ -0,0 +1,66 @@
...
+            ctx.reportIssue(equalsMethod!!, EQUALS_MESSAGE)
```
johann-beleites-sonarsource, line 60: Sorry, I just noticed this. Using `!!` in Kotlin is possible, but generally there are probably better (in the long run safer) ways of doing things. Here, the situation is slightly complicated because `equalsMethod` is a `var` and not a `val`, so the compiler can't guarantee that checking for `equalsMethod != null` still remains true during this `reportIssue` call.

Instead, what I would recommend is replace the `if-else` pattern with null-safe chaining using let:
```
        equalsMethod?.let { ctx.reportIssue(it, EQUALS_MESSAGE) }
```
Same goes for `hashCodeMethod` obviously.

johann-beleites-sonarsource (changes requested):

johann-beleites-sonarsource (approved):
LGTM!

PR #221, opened 2022-11-09
Fix javadoc error on ci about error: unknown tag

alban-auzeill:


dorian-burihabwa-sonarsource (approved):
As weird as it looks, it does the trick 🤷🏿

PR #225, opened 2022-11-10
SONARKT-243 Update external linters rules (Detekt, AndroidLint)

alban-auzeill:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/externalreport/detekt/DetektRulesDefinitionTest.kt
```diff
@@ -36,7 +36,7 @@ internal class DetektRulesDefinitionTest {
         assertThat(repository!!.name()).isEqualTo("detekt")
         assertThat(repository.language()).isEqualTo("kotlin")
         assertThat(repository.isExternal).isTrue
-        assertThat(repository.rules().size).isEqualTo(203)
+        assertThat(repository.rules().size).isEqualTo(204)
```
margarita-nedzelska-sonarsource, line 39: Looks like this should be a part of a Detekt commit

margarita-nedzelska-sonarsource (approved):
Looks nice! Thanks for taking care :)

If you want to keep 2 separate commit, please make sure DetektRulesDefinition test is part of the right commit. Otherwise, build won't pass.

Alternatively, you can just squash.

I'll let you decide.

PR #226, opened 2022-11-14
[SONARKT-129] Fixed FP for abstract methods for S1206

leonardo-pilastri-sonarsource:


kotlin-checks-test-sources/src/main/kotlin/checks/EqualsOverridenWithHashCodeCheckSample.kt
```diff
@@ -61,6 +61,27 @@ class EqualsOverridenWithHashCodeCheckSample {
         }
     }
 
+
+}
+
+abstract class AmbiguousParent {
+    abstract override fun equals(other: Any?): Boolean
+}
```
johann-beleites-sonarsource, line 69: One easy test case to add in the interest of branch coverage:
```suggestion
abstract class AmbiguousParent {
    abstract override fun equals(other: Any?): Boolean
}

abstract class AmbiguousParent4 {
    abstract override fun hashCode(): Int
}
```

johann-beleites-sonarsource (changes requested):
Just one little addition for better branch coverage, then LGTM!

johann-beleites-sonarsource (approved):

PR #229, opened 2022-11-16
Enable ktlint

johann-beleites-sonarsource:


johann-beleites-sonarsource:
Pausing this effort until the next version of ktlint is released, fixing [a bug where the disabled rules property is not picked up](https://github.com/pinterest/ktlint/pull/1671).

PR #230, opened 2022-11-16
SONARKT-207 Rule S6432: Counter Mode initialization vectors should no…

irina-batinic-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CipherModeOperationCheck.kt
```diff
@@ -0,0 +1,97 @@
...
+        val gcmExpr = thirdArgument.getGCMExpression(bindingContext, secondaries)
+        val byteExpression = gcmExpr?.getByteExpression(bindingContext, secondaries)
+
+        if (firstArgument.predictRuntimeIntValue(bindingContext) == 1 && byteExpression?.isBytesInitializedFromString(bindingContext) == true) {
+            val locations =
+                secondaries.map { SecondaryLocation(kotlinFileContext.textRange(it), "Initialization vector is configured here.") }
+                    .groupBy { location -> location.textRange.start().line() }
+                    .map { it.value[it.value.size - 1] }
+
+            kotlinFileContext.reportIssue(thirdArgument, "The initialization vector is a static value.", locations)
```
margarita-nedzelska-sonarsource, line 74: I am not completely sure, but it looks like according to the ticket the issue should be raised on the `init` method:

> The rule raises on javax.crypto.Cipher#init when it matches the following criteria: 

And we usually report on a method call and we can add a third argument as a secondary location

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CipherModeOperationCheck.kt
```diff
@@ -0,0 +1,97 @@
...
+        val secondaries = mutableListOf<PsiElement>()
+
+        val gcmExpr = thirdArgument.getGCMExpression(bindingContext, secondaries)
+        val byteExpression = gcmExpr?.getByteExpression(bindingContext, secondaries)
+
+        if (firstArgument.predictRuntimeIntValue(bindingContext) == 1 && byteExpression?.isBytesInitializedFromString(bindingContext) == true) {
+            val locations =
+                secondaries.map { SecondaryLocation(kotlinFileContext.textRange(it), "Initialization vector is configured here.") }
+                    .groupBy { location -> location.textRange.start().line() }
+                    .map { it.value[it.value.size - 1] }
```
margarita-nedzelska-sonarsource, line 72: As far as I understood, here you try not to report 2 secondaries on the same line and choose the last one. I think this logic deserves a special comment and maybe a separate method.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CipherModeOperationCheck.kt
```diff
@@ -0,0 +1,97 @@
...
+
+
+private fun KtExpression.getGCMExpression(bindingContext: BindingContext, secondaries: MutableList<PsiElement>): KtExpression? =
+    predictRuntimeValueExpression(bindingContext)
+        .getCall(bindingContext)?.let {
+            if (GCM_PARAMETER_SPEC_MATCHER.matches(it, bindingContext)) {
+                secondaries.add(it.valueArguments[1].asElement())
+                return it.valueArguments[1].getArgumentExpression()
+            } else null
+        }
```
margarita-nedzelska-sonarsource, line 97: You can configure your IDE to always add an empty line at the end

margarita-nedzelska-sonarsource (changes requested):
Nice work with your first rule 👍 

A few improvements and we can be done.

margarita-nedzelska-sonarsource (approved):
Good Job 👍

PR #231, opened 2022-11-17
SONARKT-247 Potential memory leak when regex cache holds the referenc…

margarita-nedzelska-sonarsource:
…e to the BindingContext


johann-beleites-sonarsource (approved):
LGTM!

PR #234, opened 2022-11-18
SONARKT-254 Fix Memory leak in sonar-kotlin in sonarlint because glob…

margarita-nedzelska-sonarsource:
…al cache never cleaned in the companion object

### DIFF
--- a/sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/regex/RegexContext.kt
+++ b/sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/regex/RegexContext.kt
@@ -27,24 +27,36 @@ import org.sonarsource.analyzer.commons.regex.ast.FlagSet
 import org.sonarsource.analyzer.commons.regex.ast.RegexSyntaxElement
 import org.sonarsource.kotlin.plugin.KotlinFileContext
 
+class RegexCache(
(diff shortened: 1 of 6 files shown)


johann-beleites-sonarsource:
I am worried about initializing a new cache for each file we analyze - this would mean we can't benefit from cross-file caching anymore :thinking:

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/regex/RegexContext.kt
```diff
@@ -27,24 +27,36 @@ import org.sonarsource.analyzer.commons.regex.ast.FlagSet
 import org.sonarsource.analyzer.commons.regex.ast.RegexSyntaxElement
 import org.sonarsource.kotlin.plugin.KotlinFileContext
 
+class RegexCache(
+    private val globalCache: MutableMap<Pair<List<KtStringTemplateEntry>, Int>, RegexParseResult> = mutableMapOf()
+) {
+    fun add(
```
johann-beleites-sonarsource, line 33: It seems more descriptive to me to name this `computeIfAbsent` too - or maybe at least `addIfAbsent` or even `parseAndCacheIfAbsent` or something along those lines. That way the caller gets a hint that more is happening under the hood than just an `add` operation.

johann-beleites-sonarsource (approved):
Looks good! I only have some suggestions about the `add` method naming - then LGTM!

PR #238, opened 2022-11-23
SONARKT-257 KotlinSensor should not be slow when there's no files to analyze

alban-auzeill:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
```diff
@@ -178,6 +178,19 @@ internal class KotlinSensorTest : AbstractSensorTest() {
 
     @Test
     fun `Ensure compiler crashes during BindingContext generation don't crash engine`() {
+        context.setCanSkipUnchangedFiles(false)
+        executeAnalysisWithInvalidBindingContext()
+        assertThat(logTester.logs(LoggerLevel.ERROR)).containsExactly("Could not generate binding context. Proceeding without semantics.")
+    }
+
+    @Test
+    fun `BindingContext generation don't crash if there are no files to analyze`() {
```
dorian-burihabwa-sonarsource, line 187:
```suggestion
    fun `BindingContext generation does not crash when there are no files to analyze`() {
```

dorian-burihabwa-sonarsource (approved):
LGTM! I would simply change the name of the test

PR #240, opened 2022-11-23
SONARKT-132 Rule S2116: "hashCode" and "toString" should not be called on array instances

alban-auzeill:
The RSPEC PR also needs to be reviewed: https://github.com/SonarSource/rspec/pull/1414


kotlin-checks-test-sources/src/main/kotlin/checks/ArrayHashCodeAndToStringCheckSample.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    fun falseNegative() {
+        val arrayOfArray: Array<Array<String>> = arrayOf(arrayOf("A", "B"), arrayOf("C", "D"))
+        // contrary to the above "rspecSamples()" version of "arrayOfArray", here the declaration use "val" instead of "var"
+        // and the "predictReceiverExpression(...)" is using the type the initializer "T" coming from " arrayOf(vararg elements: T)"
+        // instead of the explicit "Array<Array<String>>" variable type, which is more specific in this case.
+        // Furthermore, resolving as "T" is a mistake, is should be at least "Array<T>", because of the "vararg" modifier.
+        println(arrayOfArray.contentToString())  // false-negative
+        println(arrayOfArray.contentHashCode())  // false-negative
+    }
```
johann-beleites-sonarsource, line 106: Good catch! This is actually a small bug in `determineType`. If you simply replace the line 
```
            is KtCallExpression -> getResolvedCall(bindingContext)?.getReturnType()
```
with
```
            is KtCallExpression -> getResolvedCall(bindingContext)?.resultingDescriptor?.returnType
```
it will work ([ApiExtensions.kt:485](https://github.com/SonarSource/sonar-kotlin/blob/513ab31ec6514a24621bad242b3195c00b3f16e4/sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ApiExtensions.kt#L485))! :)

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ArgumentMatcher.kt
```diff
@@ -33,29 +34,30 @@ class ArgumentMatcher(
...
+    val isVararg: Boolean = false,
 ) {
     companion object {
         val ANY = ArgumentMatcher()
     }
 
-    fun matches(descriptor: ValueParameterDescriptor) =
-        matchesName(descriptor) && matchesNullability(descriptor)
+    fun matches(descriptor: ValueParameterDescriptor) = (isVararg == (descriptor.varargElementType != null)) &&
+        matchesName(if (isVararg) descriptor.varargElementType else descriptor.type) && matchesNullability(descriptor)
```
johann-beleites-sonarsource, line 44: Minor detail: I just noticed we are doing the nullability check after the type check. For performance reasons, we could switch them.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/ArgumentMatcher.kt
```diff
@@ -33,29 +34,30 @@ class ArgumentMatcher(
     private val typeName: String? = null,
     private val nullability: TypeNullability? = null,
     private val qualified: Boolean = true,
+    val isVararg: Boolean = false,
```
johann-beleites-sonarsource, line 37: We could make this `internal`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -52,18 +55,22 @@ class FunMatcherImpl(
...
 
-    fun matches(call: Call, bindingContext: BindingContext) = matches(bindingContext.get(RESOLVED_CALL, call))
+    fun matches(call: Call, bindingContext: BindingContext) = preCheckArgumentCount(call) &&
+        matches(bindingContext.get(RESOLVED_CALL, call)?.resultingDescriptor)
 
-    fun matches(call: ResolvedCall<*>?) = matches(call?.resultingDescriptor)
+    fun matches(resolvedCall: ResolvedCall<*>?) = preCheckArgumentCount(resolvedCall?.call) &&
+        matches(resolvedCall?.resultingDescriptor)
+
+    private fun preCheckArgumentCount(call: Call?): Boolean = call == null || (call.valueArguments.size <= maxArgumentCount)
```
johann-beleites-sonarsource, line 73: You could remove the explicit return type:
```suggestion
    private fun preCheckArgumentCount(call: Call?) = call == null || (call.valueArguments.size <= maxArgumentCount)
```

johann-beleites-sonarsource (changes requested):
Nice job! I only had a couple of minor suggestions and I think we can fix the FNs as well.

johann-beleites-sonarsource (approved):
LGTM!

PR #244, opened 2022-11-29
SONARKT-134 Rule S899: Return values should not be ignored when they contain the operation status code

alban-auzeill:
The following PR also needs to be reviewed and merged [SonarSource/rspec #1436](https://github.com/SonarSource/rspec/pull/1436)


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/IgnoredOperationStatusCheck.kt
```diff
@@ -0,0 +1,69 @@
...
+        FunMatcher(qualifier = "java.util.concurrent.BlockingQueue") {
+            withNames("offer", "remove")
+        },
+    )
+
+    override fun visitFunctionCall(callExpression: KtCallExpression, resolvedCall: ResolvedCall<*>, kotlinFileContext: KotlinFileContext) {
+        if (callExpression.isUsedAsStatement(kotlinFileContext.bindingContext)) {
+            resolvedCall.resultingDescriptor?.let { resultingDescriptor ->
+                val returnType = resultingDescriptor.returnType?.simpleName() ?: "this method";
+                val message = "Do something with the \"$returnType\" value returned by \"${resultingDescriptor.name}\"."
```
margarita-nedzelska-sonarsource, line 63: You can use triple quotes here.
```suggestion
                val message = """Do something with the "$returnType" value returned by "${resultingDescriptor.name}"."""
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/IgnoredOperationStatusCheck.kt
```diff
@@ -0,0 +1,69 @@
...
+        FunMatcher(qualifier = "java.util.concurrent.Semaphore", name = "tryAcquire"),
+        FunMatcher(qualifier = "java.util.concurrent.BlockingQueue") {
+            withNames("offer", "remove")
+        },
+    )
+
+    override fun visitFunctionCall(callExpression: KtCallExpression, resolvedCall: ResolvedCall<*>, kotlinFileContext: KotlinFileContext) {
+        if (callExpression.isUsedAsStatement(kotlinFileContext.bindingContext)) {
+            resolvedCall.resultingDescriptor?.let { resultingDescriptor ->
+                val returnType = resultingDescriptor.returnType?.simpleName() ?: "this method";
```
margarita-nedzelska-sonarsource, line 62: Surprised, there's no standard method for simple name :) But wasn't able to find it either.

margarita-nedzelska-sonarsource (approved):
Looks, good!

Just one minor suggestion!

PR #245, opened 2022-11-30
Irina/sonarkt 122

irina-batinic-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,87 @@
...
+    override val functionsToVisit = listOf(PREPARE_STATEMENT_SET, RESULT_SET_GET)
+
+    override fun visitFunctionCall(
+        callExpression: KtCallExpression,
+        resolvedCall: ResolvedCall<*>,
+        kotlinFileContext: KotlinFileContext,
+    ) {
+        val bindingContext = kotlinFileContext.bindingContext
+
+        val firstArgument = callExpression.valueArguments[0].getArgumentExpression() ?: return
```
margarita-nedzelska-sonarsource, line 55: You can add argument matcher and use `!!` here

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,87 @@
...
+            kotlinFileContext.reportIssue(firstArgument, "ResultSet indices start at 1.")
+        } else if (!isResultSet) {
+            if (firstArgumentValue == 0) {
+                kotlinFileContext.reportIssue(firstArgument, "PreparedStatement indices start at 1.")
+            } else {
+                val numberOfParameters = getNumberOfParameters(bindingContext, receiver, firstArgumentValue)
+                if (firstArgumentValue > numberOfParameters) {
+                    kotlinFileContext.reportIssue(
+                        firstArgument,
+                        "This \"PreparedStatement\" " + (if (numberOfParameters == 0) "has no" else "only has $numberOfParameters") + " parameters."
```
margarita-nedzelska-sonarsource, line 71: In Kotlin you can use String interpolation
```
"""This "PreparedStatement" ${ if (numberOfParameters == 0) "has no" else "only has $numberOfParameters" } parameters."""
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,87 @@
...
+                    )
+                }
+            }
+        }
+    }
+}
+
+private fun getNumberOfParameters(bindingContext: BindingContext, receiver: KtExpression, defaultValue: Int) =
```
margarita-nedzelska-sonarsource, line 79: You can return nullable here and drop `defaultValue`. Then on the call site, you can do 
```
getNumberOfParameters(bindingContext, receiver)?.let {...}
```

margarita-nedzelska-sonarsource (changes requested):
Nice,

There are a couple of moments that you haven't taken into consideration.
I'd suggest we drop the support of `var` as it requires much more effort than this rule deserves to make it work correctly.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,96 @@
...
+        matchedFun: FunMatcherImpl,
+        kotlinFileContext: KotlinFileContext,
+    ) {
+        val bindingContext = kotlinFileContext.bindingContext
+
+        val firstArgument = callExpression.valueArguments[0].getArgumentExpression()!!
+        val receiver = callExpression.predictReceiverExpression(bindingContext) ?: return
+        val firstArgumentValue = firstArgument.predictRuntimeIntValue(bindingContext) ?: return
+
+        if (matchedFun.qualifier.equals("java.sql.ResultSet") && firstArgumentValue == 0) {
```
margarita-nedzelska-sonarsource, line 67: you can use when here:

```
when(matchedFun) {
    RESULT_SET_GET ->...
    PREPARE_STATEMENT_SET -> ...
}
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.sonarsource.kotlin.api.INT_TYPE
+import org.sonarsource.kotlin.api.predictReceiverExpression
+import org.sonarsource.kotlin.api.predictRuntimeIntValue
+import org.sonarsource.kotlin.api.predictRuntimeStringValue
+import org.sonarsource.kotlin.api.predictRuntimeValueExpression
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+private val PREPARE_STATEMENT = FunMatcher(qualifier = "java.sql.Connection", name = "prepareStatement")
+
+private val PREPARE_STATEMENT_SET = FunMatcher(qualifier = "java.sql.PreparedStatement", nameRegex = """^set.*?""".toRegex()) {
```
margarita-nedzelska-sonarsource, line 41: I'm not sure about using reluctant quantifier here. To me, possessive should do the job but be faster`^set.*+`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -43,6 +43,7 @@ class FunMatcherImpl(
     // so you just need to specify the package
     val qualifier: String? = null,
     val names: Set<String> = emptySet(),
+    private val nameRegex: Regex?,
```
margarita-nedzelska-sonarsource, line 46: It's better to have this regex accessible to keep the consistency and in case if we need this field. Also, please add the default value `null`

And I think this property deserves a comment explaining how `nameRegex` actually works. So developers won't be confused whether its "names AND nameRegex" or "names OR nameRegex"

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/api/FunMatcherTest.kt
```diff
@@ -141,6 +141,65 @@ class FunMatcherTest {
         assertThat(funMatcher.matches(ktCallExpression2, BindingContext.EMPTY)).isFalse
     }
 
+    @Test
+    fun `Match method by type and name regex`() {
+        val funMatcher = FunMatcher {
```
margarita-nedzelska-sonarsource, line 146: You can actually use `FunMatcher(nameRegex = "...")`

And I think the test case with just `nameRegex` is missing

margarita-nedzelska-sonarsource (changes requested):
Nice work! We're almost there, just a few comments to improve

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.sonarsource.kotlin.api.INT_TYPE
+import org.sonarsource.kotlin.api.predictReceiverExpression
+import org.sonarsource.kotlin.api.predictRuntimeIntValue
+import org.sonarsource.kotlin.api.predictRuntimeStringValue
+import org.sonarsource.kotlin.api.predictRuntimeValueExpression
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+private val PREPARE_STATEMENT = FunMatcher(qualifier = "java.sql.Connection", name = "prepareStatement")
+
+private val PREPARE_STATEMENT_SET = FunMatcher(qualifier = "java.sql.PreparedStatement", nameRegex = """^set.*+""".toRegex()) {
```
margarita-nedzelska-sonarsource, line 41:
```suggestion
private val PREPARE_STATEMENT_SET = FunMatcher(qualifier = "java.sql.PreparedStatement", nameRegex = "^set.*+".toRegex()) {
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,99 @@
...
+    withArguments(ArgumentMatcher(INT_TYPE), ArgumentMatcher.ANY)
+}
+
+private val RESULT_SET_GET = FunMatcher(qualifier = "java.sql.ResultSet", nameRegex = """^get.*+""".toRegex()) {
```
margarita-nedzelska-sonarsource, line 45:
```suggestion
private val RESULT_SET_GET = FunMatcher(qualifier = "java.sql.ResultSet", nameRegex = "^get.*+".toRegex()) {
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        val firstArgument = callExpression.valueArguments[0].getArgumentExpression()!!
+        val receiver = callExpression.predictReceiverExpression(bindingContext) ?: return
+        val firstArgumentValue = firstArgument.predictRuntimeIntValue(bindingContext) ?: return
+
+        when (matchedFun) {
+            RESULT_SET_GET -> {
+                if (firstArgumentValue == 0) kotlinFileContext.reportIssue(firstArgument, "ResultSet indices start at 1.")
+            }
+
```
margarita-nedzelska-sonarsource, line 71:
```suggestion
            RESULT_SET_GET ->
                if (firstArgumentValue == 0) kotlinFileContext.reportIssue(firstArgument, "ResultSet indices start at 1.")

```

margarita-nedzelska-sonarsource (approved):
Looks very good! Nice work 👍 

A couple of small style issues. Apart from that, can be merged

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/PreparedStatementAndResultSetCheck.kt
```diff
@@ -0,0 +1,98 @@
...
+    ) {
+        val bindingContext = kotlinFileContext.bindingContext
+
+        val firstArgument = callExpression.valueArguments[0].getArgumentExpression()!!
+        val receiver = callExpression.predictReceiverExpression(bindingContext) ?: return
+        val firstArgumentValue = firstArgument.predictRuntimeIntValue(bindingContext) ?: return
+
+        when (matchedFun) {
+            RESULT_SET_GET ->
+                if (firstArgumentValue == 0) kotlinFileContext.reportIssue(firstArgument, "ResultSet indices start at 1.")
```
nicolas-gouteux-sonarsource, line 69: (I was just reading this code by interest as I need to ramp up in kotlin)...
I would rather code if (firstArgumentValue <= 0) unless you already checked for negative elsewhere...

Wohops: Hey @nicolas-gouteux-sonarsource ,

We merged this PR two weeks ago and released the Kotlin analyzer already. For now, we are not going to change anything here. The Java/Kotlin bubble is also currently focusing on other analyzers and so it is likely that we will just forget about your comment before our next sprint on Kotlin.

Can I ask you to open a PR with your suggestion so we can keep track of it, and include its review in our next sprint?

Thanks,
Michael

PR #247, opened 2022-11-30
SONARKT-138 Add rule S3981 for collections/arrays sizes comparison.

leonardo-pilastri-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CollectionSizeAndArrayLengthCheck.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        if (testedExpr is KtDotQualifiedExpression && MATCHERS.any { it.matches(testedExpr.getResolvedCall(ctx)) }) {
+            if (opToken == opWithEq) {
+                fileCtx.reportIssue(bet, ISSUE_MESSAGE_SIZE_ALWAYS_GTEQ)
+            } else if (opToken == opWithoutEq || (integerValue.isNegative() && opToken == KtTokens.EQEQ)) {
+                fileCtx.reportIssue(bet, ISSUE_MESSAGE_SIZE_NEVER_LT)
+            }
+        }
+
+    }
```
johann-beleites-sonarsource, line 71: I find this a little confusing because there is a lot happening that is, in fact, all closely related. We first have a check about whether both sides are <=0, then we assign the fact that left is <=0 to a val, then we use that val to perform 4 different if-else operations. In fact, there are two different execution streams but the way it currently is written means they are extremely interleaved with each other. I think here we can take a few more lines of code to be more explicit about these execution flows. For instance, something along these lines: 
```suggestion

        if (leftIntValue.isZeroOrNegative()) {
            checkConditionsAndSelectMessage(ctx, opToken, bet.right, leftIntValue, KtTokens.LTEQ, KtTokens.GT)
        } else if (rightIntValue.isZeroOrNegative()) {
            checkConditionsAndSelectMessage(ctx, opToken, bet.left, rightIntValue, KtTokens.GTEQ, KtTokens.LT)
        } else {
            return
        }?.let {
            fileCtx.reportIssue(bet, it)
        }
    }

    private fun checkConditionsAndSelectMessage(
        ctx: BindingContext,
        opToken: IElementType,
        testedExpr: KtExpression?,
        integerValue: Int?,
        opWithEq: KtSingleValueToken,
        opWithoutEq: KtSingleValueToken,
    ): String? {
        if (testedExpr is KtDotQualifiedExpression && MATCHERS.any { it.matches(testedExpr.getResolvedCall(ctx)) }) {
            if (opToken == opWithEq) {
                return ISSUE_MESSAGE_SIZE_ALWAYS_GTEQ
            } else if (opToken == opWithoutEq || (integerValue.isNegative() && opToken == KtTokens.EQEQ)) {
                return ISSUE_MESSAGE_SIZE_NEVER_LT
            }
        }
        return null
    }
```

sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S3981.html
```diff
@@ -0,0 +1,20 @@
+<p>The size of a collection and the length of an array are always greater than or equal to zero. So testing that a size or length is greater than or
+equal to zero doesn’t make sense, since the result is always <code>true</code>. Similarly testing that it is less than zero will always return
+<code>false</code>. Perhaps the intent was to check the non-emptiness of the collection or array instead.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+if (myList.size() &gt;= 0) { ... }
```
johann-beleites-sonarsource, line 6: Before merging, we need to update the metadata again

johann-beleites-sonarsource (changes requested):
I think we can do a little improvement in terms of how understandable/maintainable the check's code is. Otherwise, as discussed, the approach seems like a good compromise between sticking with a clean implementation and over-thinking efficiency

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/CollectionSizeAndArrayLengthCheck.kt
```diff
@@ -0,0 +1,97 @@
...
+        val leftIntValue = bet.left?.predictRuntimeIntValue(ctx)
+        val rightIntValue = bet.right?.predictRuntimeIntValue(ctx)
+
+        val msg = if(leftIntValue.isZeroOrNegative()){
+            checkConditionsAndSelectMessage(ctx, opToken, bet.right, leftIntValue, KtTokens.LTEQ, KtTokens.GT)
+        }else if(rightIntValue.isZeroOrNegative()){
+            checkConditionsAndSelectMessage(ctx, opToken, bet.left, rightIntValue, KtTokens.GTEQ, KtTokens.LT)
+        }else{
+            null
+        }
```
johann-beleites-sonarsource, line 65: Formatting: some spaces are missing here :).

johann-beleites-sonarsource (approved):
Looks good! Small formatting issue, then LGTM!

PR #248, opened 2022-11-30
SONARKT-141 Rule S3958: Intermediate Sequence/Stream methods should not be left unused

alban-auzeill:
Related rspec PR to merge: https://github.com/SonarSource/rspec/pull/356


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/StreamNotConsumedCheck.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.resolve.bindingContextUtil.isUsedAsStatement
+import org.jetbrains.kotlin.resolve.calls.model.ResolvedCall
+import org.sonar.check.Rule
+import org.sonarsource.kotlin.api.CallAbstractCheck
+import org.sonarsource.kotlin.api.FunMatcher
+import org.sonarsource.kotlin.api.FunMatcherImpl
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+private const val MESSAGE = "Refactor the code so this stream pipeline is used."
```
johann-beleites-sonarsource, line 31: We should be talking about "Sequence or Stream" pipelines here, not only stream. Or the specific one, depending on the case, of course.

johann-beleites-sonarsource (commented):
Looks good! One little comment on the message, which was not adapted from the Java message, I believe.

johann-beleites-sonarsource (approved):

PR #250, opened 2022-12-05
SONARKT-130 S2097 "equals(Any?)" should test argument type

irina-batinic-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsArgumentTypeCheck.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        var isContained = false
+        run lit@{
+            binaryExpressions.forEach { binaryExpression ->
+                val left = binaryExpression.left!!.collectDescendantsOfType<KtNameReferenceExpression>().map { it.getReferencedName() }
+                val right = binaryExpression.right!!.collectDescendantsOfType<KtNameReferenceExpression>().map { it.getReferencedName() }
+
+                isContained =
+                    (left.contains(parameter.name) && (right.contains(klassName) || right.contains(THIS_KEYWORD.value))|| right.contains("javaClass"))
+                        ((left.contains(klassName) || left.contains(THIS_KEYWORD.value) || left.contains("javaClass")) && right.contains(parameter.name))
```
johann-beleites-sonarsource, line 74: I think this expression does not do what we expect. If I comment out / remove line 74, the tests still pass, which should not be the case after changing so much logic. In fact, we never use the boolean expression on line 74, because is not linked to line 73 with any operator. I imagine this is not what you intended - I am wondering why we don't see any test failures, though.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsArgumentTypeCheck.kt
```diff
@@ -0,0 +1,85 @@
...
+                .flatMap { it.typeReference!!.text.split(".") } // class.childClass case
+                .any { it -> klassNames.contains(it) }
+
+
+        val binaryExpressions =
+            equalsMethod.collectDescendantsOfType<KtBinaryExpression> { it.operationToken == KtTokens.EQEQ || it.operationToken == KtTokens.EXCLEQ }
+
+        var isContained = false
+        run lit@{
+            binaryExpressions.forEach { binaryExpression ->
```
johann-beleites-sonarsource, line 68: You can use `binaryExpression.any {` to do this more nicely, removing the need for a `var` that you assign from within the lambda, as well as the `run`. I'd be happy to go over it together with you.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsArgumentTypeCheck.kt
```diff
@@ -0,0 +1,85 @@
...
+                isContained =
+                    (left.contains(parameter.name) && (right.contains(klassName) || right.contains(THIS_KEYWORD.value))|| right.contains("javaClass"))
+                        ((left.contains(klassName) || left.contains(THIS_KEYWORD.value) || left.contains("javaClass")) && right.contains(parameter.name))
+                if (isContained)
+                    return@lit
+            }
+        }
+
+        equalsMethod.let {
+            if (!it.isAbstract() && !hasIsExpression && !isContained) ctx.reportIssue(it, "Add a type test to this method.")
```
johann-beleites-sonarsource, line 81: We should try to return from the rule eagerly when we see that a condition to report the issue is not met. E.g., we can check `.isAbstract` before doing most of the work in this rule and avoid doing any of it. Same for `hasIsExpression`, we actually know when this is true that we can abort the computation of `isContained`.

johann-beleites-sonarsource (changes requested):
A few suggestions - happy to help go over some of the Kotlin and compiler specificities with you! :)

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsArgumentTypeCheck.kt
```diff
@@ -0,0 +1,106 @@
...
+        val right = binaryExpression.right!!.collectDescendantsOfType<KtNameReferenceExpression>()
+            .filterNot { (it.parent is KtDotQualifiedExpression || it.parent is KtSafeQualifiedExpression) && it.getReferencedName() == "javaClass" }
+            .map { it.getReferencedName() }
+        return (left.contains(parameter.name) && isContainedInList(right, klass)) || (isContainedInList(left, klass) && right.contains(
+            parameter.name
+        ))
+    }
+
+    private fun isContainedInList(list: List<String>, klass: KtClass) =
+        list.contains(klass.name) || list.contains(THIS_KEYWORD.value) || list.contains("javaClass")
```
margarita-nedzelska-sonarsource: how about naming this method `isCheckedAgainstRightType` or sth like this?

It's up to you and style, you can use a bit more functional approach `listOf(klass.name, THIS_KEYWORD.value, "javaClass").any { list.contains(it) }`

margarita-nedzelska-sonarsource (approved):
Very nice work on a difficult rule! 👍

johann-beleites-sonarsource (commented):
Comment to unblock merge.

johann-beleites-sonarsource (approved):
Approving based on @margarita-nedzelska-sonarsource 's review to unblock merge

PR #263, opened 2023-02-13
Db/sonarkt 267

dorian-burihabwa-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -89,22 +89,22 @@ class KotlinSensor(
...
         )
 
-        val filesToAnalyze = fileSystem.inputFiles(mainFilePredicate).let { mainFiles ->
+        val filePartition: Pair<Iterable<InputFile>, Iterable<InputFile>> = fileSystem.inputFiles(mainFilePredicate).let { mainFiles ->
             if (canSkipUnchangedFiles(sensorContext)) {
                 LOG.debug("The Kotlin analyzer is running in a context where it can skip unchanged files.")
                 var totalFiles = 0
-                mainFiles
-                    .filter {
-                        totalFiles++
```
johann-beleites-sonarsource, line 98: It's kind of nice to have both parts of the partition now, as we can use `.size` on both parts and add them up instead of the ugly side-effect logic of incrementing a counter.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -89,22 +89,22 @@ class KotlinSensor(
...
-                        it.status() != InputFile.Status.SAME
-                    }.also {
-                        LOG.info("Only analyzing ${it.size} changed Kotlin files out of ${totalFiles}.")
-                    }
+                val partition = mainFiles.partition {
+                    totalFiles++
+                    it.status() == InputFile.Status.SAME && reuseCPDTokens(it, sensorContext)
+                }
+                LOG.info("Only analyzing ${partition.second.size} changed Kotlin files out of ${totalFiles}.")
```
johann-beleites-sonarsource, line 100: We missed this info-level log call last time - we should use a supplier instead of always building the string

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -89,22 +89,22 @@ class KotlinSensor(
...
+                partition
```
johann-beleites-sonarsource, line 101: I would recommend using `.also` to log, as this is more explicit about the semantic dependencies in the code. It also removes the need for the expression statement to return `partition`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -89,22 +89,22 @@ class KotlinSensor(
...
             } else {
                 LOG.debug("The Kotlin analyzer is running in a context where unchanged files cannot be skipped.")
-                mainFiles
+                Pair(emptyList(), mainFiles)
```
johann-beleites-sonarsource, line 104: In Kotlin the idiomatic way of instantiating a `Pair` is using ` to `. E.g. `1 to 2` is equivalent to `Pair(1, 2)`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -89,22 +89,22 @@ class KotlinSensor(
...
             }
         }
+        val filesToAnalyze = filePartition.second
```
johann-beleites-sonarsource, line 107: For Pairs, Kotlin supports destructuring. So instead of explicitly calling `.first` or `.second`, use e.g.
```
val (_, filesToAnalyze) = filePartition
```
Even better, you don't need the intermediary val `filePartition` at all. In fact, do we even need the partition at all? I don't see where the files that are not being analyzed are being used.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -127,6 +127,27 @@ class KotlinSensor(
...
+            tokensToReuse.save()
+            try {
+                cpdTokensCache.copyForNextAnalysis(inputFile)
+                true
+            } catch (_: IllegalArgumentException) {
+                false
+            }
+        } else {
+            false
+        }
```
johann-beleites-sonarsource, line 148: This construct of checking for `null` and acting correspondingly is a typical candidate for a `?.let` construction in Kotlin. The advantage is that you save some boilerplate code and the semantic dependencies in the code are more explicit.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -127,6 +127,27 @@ class KotlinSensor(
...
+    private fun reuseCPDTokens(inputFile: InputFile,sensorContext: SensorContext): Boolean {
+        if (!sensorContext.isCacheEnabled) {
+            return false
+        }
+        val cpdTokensCache = CPDTokensCache(sensorContext.previousCache(), sensorContext.nextCache())
+        val previousTokens: List<CPDToken>? = cpdTokensCache.loadFromPreviousAnalysis(inputFile)
+        return if (previousTokens != null) {
+            val tokensToReuse = sensorContext.newCpdTokens().onFile(inputFile)
+            previousTokens.forEach { tokensToReuse.addToken(it.range, it.text) }
+            tokensToReuse.save()
```
johann-beleites-sonarsource, line 139: Another nice place where we could use a scoping function, e.g. `.apply`. Instead of saving the `tokensToReuse` to a new val and using it later on, we can restrict the scope of where the `tokensToReuse` is available and at the same time be more explicit about where side-effects happen, which involve this API. Something like:
```
            sensorContext.newCpdTokens().onFile(inputFile).apply {
                previousTokens.forEach { addToken(it.range, it.text) }
                save()
            }
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -127,6 +127,27 @@ class KotlinSensor(
...
+            try {
+                cpdTokensCache.copyForNextAnalysis(inputFile)
+                true
+            } catch (_: IllegalArgumentException) {
+                false
```
johann-beleites-sonarsource, line 144: We don't use cached tokens for this analysis if we can't cache them for the next analysis? Not saying this is the wrong decision, I'm curious about the background

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/DummyReadCache.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.kotlin
+
+import org.sonar.api.batch.sensor.cache.ReadCache
+import java.io.ByteArrayInputStream
+import java.io.InputStream
+
+class DummyReadCache(cache: Map<String, ByteArray>): ReadCache {
+    private val cache = cache
```
johann-beleites-sonarsource, line 27: I think some left-overs from copy+pasting Java code: we can directly declare the field in the constructor.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/CPDTokensCacheTest.kt
```diff
@@ -0,0 +1,241 @@
...
+import org.sonarsource.kotlin.DummyInputFile
+import org.sonarsource.kotlin.DummyReadCache
+import org.sonarsource.kotlin.DummyWriteCache
+import java.lang.IllegalArgumentException
+import java.nio.file.Path
+
+class CPDTokensCacheTest {
+    @JvmField
+    @RegisterExtension
+    var logTester = LogTesterJUnit5()
```
johann-beleites-sonarsource, line 39:
```suggestion
    val logTester = LogTesterJUnit5()
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/CPDTokensCache.kt
```diff
@@ -0,0 +1,165 @@
...
+    override fun compareTo(other: TextPointer?): Int {
+        return if (this.line == other!!.line()) {
+            Integer.compare(lineOffset, other.lineOffset())
+        } else Integer.compare(this.line, other.line())
+    }
+
+    override fun line() = line
+
+    override fun lineOffset() = lineOffset
+}
```
johann-beleites-sonarsource, line 165: Later EDIT: I think we don't actually need to implement TextRange ourselves. We have an InputFile available and should use it's API instead. Let's discuss more.

johann-beleites-sonarsource (changes requested):
As discussed, let's talk about details more tomorrow!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
dorian-burihabwa-sonarsource: Add a log message to warn of an issue that does not make the analysis fail

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -127,6 +130,27 @@ class KotlinSensor(
...
+                previousTokens.forEach { addToken(it.range, it.text) }
+                save()
+            }
+            val nextCache = sensorContext.nextCache()
+            try {
+                nextCache.copyCPDTokensFromPrevious(inputFile)
+                true
+            } catch (_: IllegalArgumentException) {
+                LOG.trace("Unable to save the CPD tokens of file ${inputFile} for the next analysis.")
+                false
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: That makes sense, I will get this fixed

leonardo-pilastri-sonarsource (approved):

leonardo-pilastri-sonarsource (approved):
Looks goot to me, nice work!

johann-beleites-sonarsource (approved):

PR #265, opened 2023-02-16
SONARKT-271 Introduce InputFile content hash cache for PR analysis

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
Thanks for all the progress.
I think there are a few things we should pay attention to:
- Not relying on state too much but passing values whenever we can
- Formatting

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+            write(key, getHash(inputFile))
+        }
+    }
+
+    fun copyFromPreviousAnalysis(inputFile: InputFile){
+        if(isEnabled()){
+            val key = contentHashKey(inputFile)
+            if(readCache!!.contains(key)){
+                writeCache!!.copyFromPrevious(key)
+                LOG.trace("Copied file $key from previous cache.")
```
johann-beleites-sonarsource, line 75: Trace-level logging => supplier

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinSensor.kt
```diff
@@ -138,6 +139,23 @@ class KotlinSensor(
...
 
+    private fun fileHasChanged(inputFile: InputFile, contentHashCache: ContentHashCache): Boolean{
+        if(contentHashCache.isEnabled()){
+            val cacheContentIsDifferent = contentHashCache.hasDifferentContentCached(inputFile)
+            if(cacheContentIsDifferent){
+                LOG.debug("Cache contained a different hash for file ${inputFile.filename()}")
+                contentHashCache.writeContentHash(inputFile)
+                return true
+            }
+            contentHashCache.copyFromPreviousAnalysis(inputFile)
```
johann-beleites-sonarsource, line 150: I feel like there is too much caching-specific logic going on in the KotlinSensor here. I think we can move some of it to the ContentHashCache, which will also allow us to simplify some code there.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+            }else{
+                writeContentHash(inputFile)
```
johann-beleites-sonarsource, line 77: If I understand correctly, we should never be able to enter this case. If we move some logic from KotlinSensor to ContentHashCache, as mentioned in a comment above, I think we can avoid this.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }else{
+            LOG.trace("Cannot write to cache when disabled.")
+        }
+    }
+
+    private fun write(key: String, hash: ByteArray){
+        if(isEnabled()){
+            LOG.trace("Storing file content hash for key: $key")
```
johann-beleites-sonarsource, line 86: Trace-level log

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+                LOG.trace("Copied file $key from previous cache.")
+            }else{
+                writeContentHash(inputFile)
+            }
+        }else{
+            LOG.trace("Cannot write to cache when disabled.")
+        }
+    }
+
+    private fun write(key: String, hash: ByteArray){
```
johann-beleites-sonarsource, line 84: I am not convinced that this method here makes sense. It is only called by one other method and its body is pretty simple. Besides, we do a check here whether the cache is enabled - it would make sense to do it earlier, to avoid some useless computation. I think we can merge the two methods.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+        }else{
+            LOG.trace("Cannot write to cache when disabled.")
+        }
+    }
+
+    private fun read(key: String): ByteArray? {
+        if(enabled){
+            if(readCache!!.contains(key)){
+                val stream = readCache!!.read(key)
+                return stream.readAllBytes().also { stream.close() }
```
johann-beleites-sonarsource, line 97: Let's use "try-with-resources" here. In Kotlin, this is done via the `.use` method.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+            }else{
+                LOG.trace("Key $key was not found in ContentHashCache")
```
johann-beleites-sonarsource, line 99: Trace-level log

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }else{
+            LOG.trace("Cannot read from cache when disabled.")
+        }
+        return null
+    }
+
+    private fun getHash(inputFile: InputFile): ByteArray{
+        val inputFileBytes = inputFile.contents().byteInputStream().readAllBytes()
```
johann-beleites-sonarsource, line 108: Another good place for `.use`

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/DummyReadCache.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.kotlin
+
+import org.sonar.api.batch.sensor.cache.ReadCache
+import java.io.ByteArrayInputStream
+import java.io.InputStream
+
+class DummyReadCache(cache: Map<String, ByteArray>): ReadCache {
+    private val cache = cache
```
johann-beleites-sonarsource, line 27: Some relicts of copy+pasting Java code. We can declare the `cache` field directly in the constructor.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/DummyReadCache.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun read(key: String): InputStream {
+        if (!cache.containsKey(key)) {
+            throw IllegalArgumentException("cache")
```
johann-beleites-sonarsource, line 31: Is the messages meant to be just `"cache"`?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    private var enabled = false
+    private var readCache: ReadCache? = null
+    private var writeCache: WriteCache? = null
+    private var computedHashes = mutableMapOf<String, ByteArray>()
+
+    constructor(sensorContext: SensorContext){
+        this.enabled = sensorContext.isCacheEnabled
+        if(this.enabled){
+            LOG.info("Content Hash Cache was initialized")
```
johann-beleites-sonarsource, line 44: Is this information important enough to be printed at info level?

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/ContentHashCacheTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        assertThat(logTester.logs(LoggerLevel.TRACE)).contains("Copied file kotlin:contentHash:MD5:DummyFileChanged.kt from previous cache.")
+    }
+
+    private fun emptyContentHashCache(): ContentHashCache{
+        val readCache = DummyReadCache(mapOf())
+        val writeCache = DummyWriteCache(readCache = readCache)
+        return cacheFromContextData(readCache, writeCache, true)
+    }
+
+    private fun contentHashCacheOf( previousValue: Pair<String, ByteArray> ): ContentHashCache{
```
johann-beleites-sonarsource, line 129: How come we are using a Pair instead of simply 2 parameters here?

johann-beleites-sonarsource (changes requested):
A couple of suggestions. It is always most difficult to have multiple people review your PR :smile: .

I noticed that the use of spaces does not adhere with the Kotlin code style. Are you using Eclipse? If yes, does Eclipse autoformat Kotlin? If you are using IntelliJ, I suggest autoformatting the code regularly, it does a pretty decent job.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,81 @@
...
+        fun of(ctx: SensorContext) = if (ctx.isCacheEnabled) {
+            LOG.debug("Content hash cache was initialized")
+            ContentHashCache(ctx.previousCache(), ctx.nextCache())
+        } else {
+            LOG.debug("Content hash cache is disabled")
+            null
+        }
+    }
+
+    fun hasDifferentContentCached(inputFile: InputFile): Boolean {
```
dorian-burihabwa-sonarsource, line 46: Has this method has the side-effect a writing to the cache, we should probably add some documentation on the method

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/ContentHashCacheTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        val contentHashCacheDisabled = ContentHashCache.of(context)
+        assertThat(contentHashCacheDisabled == null).isTrue
+    }
+
+    @Test
+    fun `test NullPointerException on context with cache disabled`() {
+        context.isCacheEnabled = false
+        val contentHashCacheDisabled = ContentHashCache.of(context)
+        assertThrows<NullPointerException> { contentHashCacheDisabled!!.hasDifferentContentCached(dummyFile) }
+    }
```
dorian-burihabwa-sonarsource, line 89: I think this test is redundant with the one above since the goal is to test the behavior of `ContentHashCache.of`

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
```diff
@@ -529,6 +533,31 @@ internal class KotlinSensorTest : AbstractSensorTest() {
...
+        val checkFactory = checkFactory("S1764")
+
+        sensor(checkFactory).execute(context)
+        assertThat(logTester.logs(LoggerLevel.DEBUG))
+            .contains("Content hash cache was initialized")
+        assertThat(logTester.logs(LoggerLevel.INFO))
+            .contains("Only analyzing 2 changed Kotlin files out of 3.")
+
+
+    }
```
dorian-burihabwa-sonarsource, line 559: Should we also test a case where we explicitly test when cache is disabled and we fall back on the input file status?

dorian-burihabwa-sonarsource (changes requested):
Great job, not much to change here. Feel free to come back to me if you want to discuss my feedback

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,81 @@
...
+
+private val LOG = Loggers.get(ContentHashCache::class.java)
+private const val HASH_ALGORITHM = "MD5"
+private val messageDigest = MessageDigest.getInstance(HASH_ALGORITHM)
+private const val CONTENT_HASHES_KEY = "kotlin:contentHash:$HASH_ALGORITHM:"
+
+class ContentHashCache private constructor(private val readCache: ReadCache, private val writeCache: WriteCache) {
+
+    companion object {
+        fun of(ctx: SensorContext) = if (ctx.isCacheEnabled) {
```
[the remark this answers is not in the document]
leonardo-pilastri-sonarsource: That's a good point, I think I got caught up in the functional madness of Kotlin

dorian-burihabwa-sonarsource (approved):

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/caching/ContentHashCache.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    /**
+     * Checks if for the inputFile, there is a different content
+     * hash cached. If that is the case, or the file has never
+     * been cached before, takes care of writing a
+     * new entry to the cache. Otherwise, it will copy the content hash
+     * from the previous cache.
+     */
```
dorian-burihabwa-sonarsource, line 52:
```suggestion
    /**
     * Checks if the inputFile has a different content hash cached.
     * If that is the case, or the file has never been cached before, takes care of writing a new entry to the cache.
     * Otherwise, it will copy the content hash from the previous cache.
     */
```

dorian-burihabwa-sonarsource (approved):

johann-beleites-sonarsource (approved):
LGTM!

PR #273, opened 2023-03-02
Add missing -SNAPSHOT key to version number 

dorian-burihabwa-sonarsource:
Fix a bug where artifacts would be published without version number making it impossible to distinguish between build artifacts.

This would result in ITs in the CI the pipeline failing because the tests would run against mismatched versions of the kotlin analyzer.


irina-batinic-sonarsource (approved):
LGTM!

PR #286, opened 2023-03-22
SONARKT-283 Implement rule S6511: "When" statements should be used instead of chained "if" statements

kaufco:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.jetbrains.kotlin.KtNodeTypes
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.psi.KtContainerNodeForControlStructureBody
+import org.jetbrains.kotlin.psi.KtIfExpression
+import org.sonar.check.Rule
+import org.sonar.check.RuleProperty
+import org.sonarsource.kotlin.api.AbstractCheck
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+const val DEFAULT_THRESHOLD = 2
```
leveretka, line 31: I feel that 2 is a bit too aggressive, we might get many unhappy users there. I see ruling has a lot of issues raised there. Maybe 3 will be already much better?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+private fun isElseElement(element: PsiElement) = element.node.elementType === KtNodeTypes.ELSE
+
+private fun isIfChainLongerThanOrEqualThreshold(expression: KtIfExpression, threshold: Int): Boolean {
+    var elseBranch = expression.`else`
+    var length = 1
+
+    while (elseBranch is KtIfExpression && ++length < threshold) {
```
leveretka, line 57: Here you ignore `else` if it's not followed by `if`. This makes `if/else if/else` equal to `if/else if` in terms of this check. However, first example contains 3 branches and should better be refactored, but the second one contains only 2, and can be left as is.

Wdyt?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    @RuleProperty(
+        key = "threshold",
+        description = """Number of chained "if" statements after which should be merged.""",
+        defaultValue = DEFAULT_THRESHOLD.toString(),
+    )
+    var threshold: Int = DEFAULT_THRESHOLD
+
+    override fun visitIfExpression(expression: KtIfExpression, context: KotlinFileContext) {
+        val isHeadOfIfChain = !isElseElement(expression.parent)
```
leveretka, line 44: I think you can just inline this value and not use a separate variable

kotlin-checks-test-sources/src/main/kotlin/checks/MergeIfElseIntoWhenCheckSample.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    fun returnOneIfNoElseCompliant(value: Int): String {
+        if (value > 0) { // Complaint
+            return "positive"
+        }
+        return "negative"
+    }
+
+    fun returnTwoIfNoncompliant(value: Int): String {
+        return if (value > 0) { // Noncompliant {{Merge chained "if" statements into a single "when" statement.}}
```
leveretka, line 21: you can also check issue location with comments:

```
return if (value > 0) {
//     ^^        
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+        if (isHeadOfIfChain && isIfChainLongerThanOrEqualThreshold(expression, threshold)) {
+            context.reportIssue(expression.ifKeyword, """Merge chained "if" statements into a single "when" statement.""")
```
leveretka, line 46: I think it will also be nice to have secondary locations with other chained `if`s. Wdyt?

leveretka (changes requested):

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
kaufco: This was intended so, because I thought that we should count the number of conditions (or `if` statements), not the number resulting branches. I.e., threshold 3 hits on `if/else if/else if` and on `if/else if/else if/else`, but not on `if/else if/else. I think this is more natural, because three chained `if` statements (but without a concluding `else`) is more complex than only two `if` statements, but with an `else`.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
kaufco: I see your point, I was also a bit concerned about this. Rised to 3. Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.kotlin.checks
+
+import org.jetbrains.kotlin.KtNodeTypes
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.psi.KtContainerNodeForControlStructureBody
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
kaufco: The purpose of this local variable was documentation, but the expression may be too trivial for that. Fixed.

kotlin-checks-test-sources/src/main/kotlin/checks/MergeIfElseIntoWhenCheckSample.kt
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
kaufco: Added issue location check also for `return if` and nested if statements. Fixed.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheckTest.kt
```diff
@@ -0,0 +1,30 @@
...
+package org.sonarsource.kotlin.checks
+
+internal class MergeIfElseIntoWhenCheckTest {
+
+    class WithDefaultThreshold : CheckTest(MergeIfElseIntoWhenCheck())
+
+    class WithThreshold3 : CheckTest(
+        MergeIfElseIntoWhenCheck().apply { threshold = 3 },
+        MergeIfElseIntoWhenCheck::class.simpleName + "Threshold3Sample.kt"
+    )
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S6511.html
```diff
@@ -0,0 +1,73 @@
+<h2>Why is this an issue?</h2>
+<p>The Kotlin <code>when</code> statement is more powerful than the Java <code>switch</code> statement. It can be used to compare a selector against a
```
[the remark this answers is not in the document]
kaufco: I changed this in the RSpec and updated ruling metadata. Fixed.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheckTest.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import org.junit.jupiter.api.Test
+import org.sonarsource.kotlin.verifier.KotlinVerifier
+
+internal class MergeIfElseIntoWhenCheckTest : CheckTest(MergeIfElseIntoWhenCheck()) {
+
+    @Test
+    fun `with threshold 2`() {
+        KotlinVerifier(MergeIfElseIntoWhenCheck().apply { threshold = 2 }) {
+            this.fileName = MergeIfElseIntoWhenCheck::class.simpleName + "Threshold2Sample.kt"
```
leveretka, line 30: You can use prooperty `checkName` from `CheckTest` class here. `this.fileName = "${checkName}Threshold2Sample.kt`

kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MergeIfElseIntoWhenCheck.kt
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+
+
+
+
+
+
+
```
leveretka, line 100: Looks like too many empty lines :)

leveretka (approved):
Very nice work on the rule 👍 

Please, remove odd empty lines and you can merge

PR #292, opened 2023-03-28
Fix broken "Sonar_way_profile.json" caused by a missing comma

alban-auzeill:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinProfileDefinitionTest.kt
```diff
@@ -19,19 +19,30 @@
...
 
 internal class KotlinProfileDefinitionTest {
 
+    @Test
+    fun `Sonar_way_profile is a valid json file`() {
+        val classLoader = KotlinProfileDefinition::class.java.classLoader
+        val jsonContent = classLoader.getResource(KotlinProfileDefinition.PATH_TO_JSON)?.readText()
+        val json = Gson().fromJson(jsonContent, JsonObject::class.java)
+        assertThat(json.get("name").asString).isEqualTo("Sonar way")
+        assertThat(json.get("ruleKeys").asJsonArray).hasSizeGreaterThan(80)
```
kaufco, line 36: Maybe because we use 80 and "Sonar way" in both tests, we should declare a constant for it, so we keep both tests in sync automatically?

kaufco (changes requested):

PR #294, opened 2023-03-30
SONARKT-307 Fix FP in S1128 (Unused import) in case if delegate is a …

leveretka:
…local variable

### DIFF
--- a/sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
+++ b/sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
@@ -19,8 +19,11 @@
  */
 package org.sonarsource.kotlin.checks
 
-import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+import org.jetbrains.kotlin.descriptors.VariableAccessorDescriptor
+import org.jetbrains.kotlin.descriptors.VariableDescriptorWithAccessors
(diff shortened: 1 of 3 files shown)


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/UnnecessaryImportsCheck.kt
```diff
@@ -170,8 +171,8 @@ class UnnecessaryImportsCheck : AbstractCheck() {
                     ?.resultingDescriptor
                     ?.fqNameOrNull()?.let { listOf(it) }
                     ?: (propDelegate.parent as? KtProperty)?.let { ktProperty ->
-                        (bindingContext.get(BindingContext.DECLARATION_TO_DESCRIPTOR, ktProperty) as? PropertyDescriptor)?.let { propDescriptor ->
-                            propDescriptor
+                        (bindingContext.get(BindingContext.DECLARATION_TO_DESCRIPTOR, ktProperty) as? VariableDescriptorWithAccessors)?.let { descriptor ->
+                            descriptor
                                 .accessors
                                 .mapNotNull {
                                     bindingContext.get(BindingContext.DELEGATED_PROPERTY_RESOLVED_CALL, it)
```
kaufco, line 178: Not caused by your commit, but I think we should reduce the complexity of this function in general.

kaufco (approved):
LGTM, apart from one function that should optionally be refactored for a better understanding of the code.

PR #301, opened 2023-04-03
Implement rule S6518 : Element access should use indexed access operators

kaufco:


kotlin-checks-test-sources/src/main/kotlin/checks/IndexedAccessCheckSample.kt
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface Grid {
+    operator fun get(row: Int, column: Int): Int
+    operator fun set(row: Int, column: Int, value: Int)
+}
+
+interface Lisp<T> {
+    /*operator*/ fun get(index: Int)
+}
```
alban-auzeill, line 39: We should have some test cases about access to indexed elements of it's own class or parent class using `get(index)`, `this.get(index)`, `this@ChildClass.get(index)`, `super.get(index)`. I notice that we have a false positive about `super` and we should make our decision explicit about `this`.

Note: the rule S6519 is also impacted by operator overloading limitation, perhaps some logic related to `super` and `this` could be put in common code. The logic is [here](https://github.com/SonarSource/sonar-kotlin/blob/42ccc6518f9b01052472695d91306f167bfd84cd/sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/EqualsMethodUsageCheck.kt#L45)

```kotlin
open class ParentClass {
    private val _array = mutableListOf<String>()

    open operator fun get(index: Int): String {
        return _array[index]
    }

    open operator fun set(index: Int, value: String) {
        _array[index] = value
    }

    val size: Int get() = _array.size
}

class ChildClass : ParentClass() {

    override operator fun get(index: Int): String {
        // The following code does not compile: return super[index]
        return super.get(index) // Noncompliant but false positive, calling "get" here should be compliant
    }

    override operator fun set(index: Int, value: String) {
        // The following code does not compile: super[index] = value
        super.set(index, value) // Noncompliant but false positive, calling "set" here should be compliant
    }


    fun getOrEmpty(index: Int): String =
        if (index in 0 until size)
            get(index) // Noncompliant and it's not possible to simplify with "[index]", only with "this[index]"
                       // and I don't have a strong feeling that we should enforce this case.
        else
            ""

    fun setIfExist(index: Int, value: String) {
        if (index in 0 until size) {
            set(index, value) // Noncompliant and it's not possible to simplify with "[index] = value", only with "this[index] = value"
                              // and I don't have a strong feeling that we should enforce this case.
        }
    }

    inner class InnerClass {
        operator fun get(index: Int): String {
            return this@ChildClass.get(index) // Noncompliant but should we really enforce this delegating case?
        }

        operator fun set(index: Int, value: String) {
            this@ChildClass.set(index, value) // Noncompliant but should we really enforce this delegating case?
        }
    }
}
```

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -207,6 +214,7 @@ fun FunMatcher(
     dynamic: Boolean? = null,
     extensionFunction: Boolean? = null,
     suspending: Boolean? = null,
+    operator: Boolean? = null,
```
alban-auzeill, line 217: `operator` is a valid variable name but it could be confuse by some developers with the kotlin function modifier keyword. I would rename it with `overloadedOperator`.

kaufco: I would have called it `isOperator`, but this would break the naming schema of the existing arguments (i.e., dynamic instad isDynamic et c.) overloadedOperator, operatorFunction or something like that sounds good.

leveretka: I think that is fine to use `operator`. Also, you can use `isOperator` and refactor all other boolean properties. 
`overloadedOperator` or `operatorFunction` sound more confusing to me.

leveretka (changes requested):
Nice implementation, I think I agree with @alban-auzeill we need to cut a few FPs, other than that looks good!

kotlin-checks-test-sources/src/main/kotlin/checks/IndexedAccessCheckSample.kt
kaufco: Good point about the `super` and `this`! Thanks for the test cases. The logic should be: "if a dot expression (i.e., with an lhs receiver) uses a get / set operator function, it should be replaced with an indexed accessor". According to that, I fixed two of the three suggested corner cases:

- `super.get(index)` -> follows the logic, but `super[index]` is not allowed, therefore FP. Fixed.
- `get(index)` -> Does not follow the logic, because this is no dot expression, therefore FP. I.e., `[index]` is not allowed, while `this[index]` is not a simplification of the original expression. Fixed.
- `return this@ChildClass.get(index)` -> does follow the logic and `return this@ChildClass[index]` is a simplification. Therefore not an FP in my eyes. Not fixed.

Would you agree with that?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
kaufco: I opted for the refactoring. Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -152,6 +155,9 @@ class FunMatcherImpl(
 
     private fun checkIsSuspending(descriptor: CallableDescriptor) =
         suspending?.let { it == descriptor.isSuspend } ?: true
+
+    private fun checkIsOperator(descriptor: CallableDescriptor) =
+        operator?.let { it == ((descriptor as? FunctionDescriptor)?.isOperator ?: false) } ?: true
```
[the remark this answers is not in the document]
kaufco: Fixed.

PR #304, opened 2023-04-11
SONARKT-287 FP S6512 related to setter visibility, overrides and annotations

alban-auzeill:


kotlin-checks-test-sources/src/main/kotlin/checks/PropertyGetterAndSetterUsageCheckSample.kt
```diff
@@ -19,14 +21,172 @@ open class PropertyGetterAndSetterUsageCheckSample {
...
+        protected fun getF8(): Int = f8 // Noncompliant
+        private fun setF8(value: Int) { f8 = value }
+
+        // private getter / private setter
+        private var f9: Int = 0
+        private fun getF9(): Int = f9 // Noncompliant
+        private fun setF9(value: Int) { f9 = value }
+
+        // ______________________________________________
+        // 2) Setter with visibility > getter visibility
```
johann-beleites-sonarsource, line 78: I would still expect the getter in these examples to use the property access syntax here, so for the `f10` example, to look like something like this:
```
        internal var f10: Int = 0
        public fun setF10(value: Int) { f10 = value }
```

johann-beleites-sonarsource, line 78: I would still expect the getter in these examples to use the property access syntax here, so for the `f10` example, to look like something like this:
```
        internal var f10: Int = 0
        public fun setF10(value: Int) { f10 = value }
```
I've also added a couple of other comments for some of the examples.
WDYT?

kotlin-checks-test-sources/src/main/kotlin/checks/PropertyGetterAndSetterUsageCheckSample.kt
```diff
@@ -19,14 +21,172 @@ open class PropertyGetterAndSetterUsageCheckSample {
...
+        private var f21: Int = 0
+        internal fun setF21(value: Int) { f21 = value } // Compliant
+
+        // missing getter / protected setter
+        private var f22: Int = 0
+        protected fun setF22(value: Int) { f22 = value } // Compliant
+
+        // missing getter / private setter
+        private var f23: Int = 0
+        private fun setF23(value: Int) { f23 = value } // Compliant
```
johann-beleites-sonarsource, line 129: The field here has the same visibility as the setter - that means we don't need the setter.

kotlin-checks-test-sources/src/main/kotlin/checks/PropertyGetterAndSetterUsageCheckSample.kt
```diff
@@ -19,14 +21,172 @@ open class PropertyGetterAndSetterUsageCheckSample {
...
+        internal fun setF13(value: Int) { f13 = value }
+
+        // private getter / public setter
+        private var f14: Int = 0
+        private fun getF14(): Int = f14 // Compliant
+        public fun setF14(value: Int) { f14 = value }
+
+        // private getter / internal setter
+        private var f15: Int = 0
+        private fun getF15(): Int = f15 // Compliant
```
johann-beleites-sonarsource, line 107: The getter has the same visibility as the property -> we don't need it.

johann-beleites-sonarsource (commented):
I am not 100% convinced of the logic change yet, however, it feels like I may be missing something here.

kotlin-checks-test-sources/src/main/kotlin/checks/PropertyGetterAndSetterUsageCheckSample.kt
alban-auzeill: It's not equivalant, your proposal expose an implicit `internal set(value) { field = value }` property setter. Having two setters could be dangerous if there's a specific logic that must be called in `setF10`. The smallest equivalent replacement could be:

- Before

```kotlin
    private var f10: Int = 0
    internal fun getF10(): Int = /* ... getter logic */ f10
    public fun setF10(value: Int) { /* ... setter logic */ f10 = value }
```

- After

```kotlin
    internal var f10: Int = 0
        get() = /* getter logic */ field
        private set // hide the implicit setter
    public fun setF10(value: Int) { /* ... setter logic */ f10 = value }
```

So for now my fear is to ask a developer to change the code without enough assistance about the implicit setter visibility, and this could become a bug. I prefer to not raise an issue with the current sate of the rule description, and when we will work again to improve this rule (see [SONARKT-308](https://sonarsource.atlassian.net/jira/software/c/projects/SONARKT/issues/SONARKT-308)), we could improve the educational section about implicit getter and setter.

kotlin-checks-test-sources/src/main/kotlin/checks/PropertyGetterAndSetterUsageCheckSample.kt
alban-auzeill: It will be done later by [SONARKT-308](https://sonarsource.atlassian.net/jira/software/c/projects/SONARKT/issues/SONARKT-308)

leonardo-pilastri-sonarsource (approved):
Looks good! Maybe you can think about making the `KtNamedFunction.isLessVisibleThan` method as a public extension function, you decide if it's worth it!

alban-auzeill:
> Maybe you can think about making the `KtNamedFunction.isLessVisibleThan` method as a public extension function

I didn't find any other code dealing with all kinds of visibility (`isPublic`, `isProtected`, `isInternal`, `isPrivate`)
So putting the code in a common place would be premature. And the logic of `isLessVisibleThan` could be only valid for this rule, because the visibility is not just:

-  `public > internal > protected > private`

but

-  `public > (internal != protected) > private`
   because `internal` visibility does not enclose `protected` visibility and vice versa.

PR #306, opened 2023-04-17
SONARKT-251 Fix FP in rule S1186 when @Suppress("DEPRECATION") is used

leveretka:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/verifier/KtTestChecksVisitor.kt
```diff
@@ -28,7 +31,14 @@ import org.sonarsource.kotlin.visiting.KotlinFileVisitor
...
 class KtTestChecksVisitor(private val checks: List<AbstractCheck>) : KotlinFileVisitor() {
     init {
-        checks.forEach { it.initialize(RuleKey.of("Kotlin", "Dummy")) }
+        checks.forEach { check ->
+            val key = (check.javaClass.annotations.findAnnotation(FqName("org.sonar.check.Rule"))
+                ?.arguments
+                ?.find { it.name?.asString() == "key" } as? ReflectJavaLiteralAnnotationArgument)
+                ?.value as? String
+
+            check.initialize(RuleKey.of("kotlin", key ?: "Dummy"))
```
johann-beleites-sonarsource, line 40: Maybe we could even assert that the key is never `null` and remove the "Dummy" fallback? Is there any situation where `key == null` would be expected behavior?

johann-beleites-sonarsource (commented):
Nice, finally proper rule keys in the tests! :partying_face:

johann-beleites-sonarsource (approved):

PR #312, opened 2023-04-19
SONARKT-292 Implement rule S6515: Singleton pattern should use object declarations or expressions

kaufco:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+            klass.fqName?.let {
+                singletonClassCandidates.add(it.toString())
+            }
+        }
+    }
+}
+
+private fun isSingletonClassCandidate(klass: KtClass): Boolean {
+    val constructors = klass.allConstructors
+    return klass.isPrivate() || (constructors.isNotEmpty() && constructors.all(KtModifierListOwner::isPrivate))
```
johann-beleites-sonarsource, line 83: From Java experience I have the feeling that singletons usually only declare one no-arg constructor (or none, although that case may be harder to handle, as we can't find if it is instantiated elsewhere). So maybe we can reduce our scope here a little and ignore classes with multiple constructors (why have them if one isn't called?). Also, why pass arguments to a constructor if they are always the same? This can also help us later to determine the number of invocations more efficiently. I am happy to discuss this point though, perhaps it is different in Kotlin (I am not sure).

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+    override fun visitCallExpression(expression: KtCallExpression) {
+        val fqName = getCalledConstructorClassFqNameOrNull(expression) ?: return
+        if (!singletonClassCandidates.contains(fqName)) return
+        if (singleConstructorCallByClass.put(fqName, expression) != null) {
+            singleConstructorCallByClass.remove(fqName)
+            singletonClassCandidates.remove(fqName)
+        }
+    }
+
+    private fun getCalledConstructorClassFqNameOrNull(expression: KtCallExpression): String? {
```
johann-beleites-sonarsource, line 104: Instead of using this custom logic, you can use a `FunMatcher` that matches the constructor you are looking for. `FunMatcher`s also try to run the logic in an optimized way, e.g. checking the number of arguments before resolving too much using semantics.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+    if (bodyExpression.statements.last() != callExpression) return false
+    val call = (((bodyExpression.parent as? KtFunctionLiteral)
+        ?.parent as? KtLambdaExpression)
+        ?.parent as? KtLambdaArgument)
+        ?.parent as? KtCallExpression ?: return false
+
+    val property = (call.parent as? KtPropertyDelegate)?.parent as? KtProperty ?: return false
+    if (call.valueArguments.size != 1 || !isDeclaredInCompanionObject(property)) return false
+    val calleeReference = call.calleeExpression as? KtReferenceExpression ?: return false
+    return bindingContext[BindingContext.REFERENCE_TARGET, calleeReference]?.name?.toString() == "lazy"
```
johann-beleites-sonarsource, line 127: This logic should be replaceable by some generic logic we already have. For one, there is `getResolvedCall(BindingContext)` defined on e.g. the call expression (amongst others). Even better, you can use a FunMatcher here as well.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+}
+
+private fun isDeclaredInCompanionObject(property: KtProperty): Boolean =
```
johann-beleites-sonarsource, line 130: What about singleton instance fields declared on file level? Do we ignore those on purpose?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+}
+
+private fun isInitializingCall(callExpression: KtCallExpression): Boolean {
+    val property = callExpression.parent as? KtProperty ?: return false
+    return isDeclaredInCompanionObject(property)
+}
+
+private fun isLazyInitializingCall(callExpression: KtCallExpression, bindingContext: BindingContext): Boolean {
+    val bodyExpression = callExpression.parent as? KtBlockExpression ?: return false
+    if (bodyExpression.statements.last() != callExpression) return false
```
johann-beleites-sonarsource, line 118: You can probably use `!==` for a more efficient instance equality here.

johann-beleites-sonarsource (changes requested):

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
kaufco: Unfortunately yes, because a TLD instance field implies that the constructor is not private. But thinking about it again: if the constructor is not private, but the class is, we could do the check also for TLD instance fields. I'll add that.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+        if (singleConstructorCallByClass.put(fqName, expression) != null) {
+            singleConstructorCallByClass.remove(fqName)
+            singletonClassCandidates.remove(fqName)
+        }
+    }
+
+    private fun getCalledConstructorClassFqNameOrNull(expression: KtCallExpression): String? {
+        val calleeReference = expression.calleeExpression as? KtReferenceExpression ?: return null
+        val calledConstructor = bindingContext[BindingContext.REFERENCE_TARGET, calleeReference] as? ConstructorDescriptor
+        return calledConstructor?.constructedClass?.fqNameSafe?.toString()
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+}
+
+private fun isSingletonClassCandidate(klass: KtClass): Boolean {
+    val constructors = klass.allConstructors
+    return klass.isPrivate() || (constructors.isNotEmpty() && constructors.all(KtModifierListOwner::isPrivate))
+}
+
+private fun KtModifierListOwner.isPrivate(): Boolean = visibilityModifier()?.text == "private"
```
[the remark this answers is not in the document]
kaufco: Ah, nice, didn't see that. Thanks :-) 

Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
kaufco: Fixed. Didn't use a FunMatcher though, because here I do not want to match the function, but get the constructor type name, so I can differentiate between three states of the class:

- instantiated never
- instantiated exactly once
- instantiated more than once

I do realize though that here is a bit code duplication with the function matcher, since it determines the constructed class name in exactly the same way. Should we leverage this and create some helper functions here that could be reused, wdyt?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -39,10 +39,14 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.overriddenTreeUniqueAsSequenc
 
 private const val VARARG_PREFIX = "vararg ";
 
+private fun Set<String>.addOptional(optionalString: String?): Set<String> =
```
johann-beleites-sonarsource, line 42: How about calling the method `addIfNonEmpty`, making it clear on which condition it will be added?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,132 @@
...
+            klass.fqName?.let {
+                singletonClassCandidates.add(it.toString())
+            }
+        }
+    }
+}
+
+private fun isSingletonClassCandidate(klass: KtClass): Boolean {
+    val constructors = klass.allConstructors
+    return klass.isPrivate() || (constructors.isNotEmpty() && constructors.all(KtModifierListOwner::isPrivate))
```
johann-beleites-sonarsource, line 85: I would probably write 
```
constructors.all { it.isPrivate() }
```
instead of passing a function reference, out of readability reasons.

However, I don't think we have a clearly defined rule here and this is my personal taste. Seeing this, maybe it's a good topic for a coffee break one day.

johann-beleites-sonarsource (changes requested):
It is hard to predict the TP/FP rate and runtime performance of a rule such as this one, so let's make sure we pay extra attention during verification!

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
kaufco: Fixed, since I know now that the compiler generates the same code for both.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SingletonPatternCheck.kt
```diff
@@ -0,0 +1,132 @@
...
+    if (bodyExpression.statements.last() !== callExpression) return false
+    val call = (((bodyExpression.parent as? KtFunctionLiteral)
+        ?.parent as? KtLambdaExpression)
+        ?.parent as? KtLambdaArgument)
+        ?.parent as? KtCallExpression ?: return false
+
+    val property = (call.parent as? KtPropertyDelegate)?.parent as? KtProperty ?: return false
+    if (call.valueArguments.size != 1 || !isStaticProperty(property)) return false
+    val calleeReference = call.calleeExpression as? KtReferenceExpression ?: return false
+    return bindingContext[BindingContext.REFERENCE_TARGET, calleeReference]?.name?.toString() == "lazy"
```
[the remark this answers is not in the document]
kaufco: Yes, makes sense. I somehow overlooked this. Fixed.

johann-beleites-sonarsource (approved):
LGTM!

PR #314, opened 2023-04-21
SONARKT-119  Rule S6207: Redundant functions should be avoided in data classes

irina-batinic-sonarsource:


kotlin-checks-test-sources/src/main/kotlin/checks/RedundantMethodsInDataClassesCheckSample.kt
```diff
@@ -0,0 +1,104 @@
...
+
+class RedundantMethodsInDataClassesCheckSample {
+
+    data class Person1(val name: String, val age: Int) {
+        override fun equals(other: Any?): Boolean { // Noncompliant {{Remove this redundant method which is the same as a default one.}}
+//                   ^^^^^^
+            return other is Person1 && other.name == name && other.age == age
+        }
+
+        override fun hashCode() = Objects.hash(name, age) // Noncompliant {{Remove this redundant method which is the same as a default one.}}
```
leveretka, line 13: Actually, if you look at the bytecode, it's not what is actually generated. 

```
// access flags 0x1
  public hashCode()I
    ALOAD 0
    GETFIELD checks/RedundantMethodsInDataClassesCheckSample$Person2.name : Ljava/lang/String;
    DUP
    IFNULL L0
    INVOKEVIRTUAL java/lang/Object.hashCode ()I
    GOTO L1
   L0
    POP
    ICONST_0
   L1
    BIPUSH 31
    IMUL
    ALOAD 0
    GETFIELD checks/RedundantMethodsInDataClassesCheckSample$Person2.age : I
    INVOKESTATIC java/lang/Integer.hashCode (I)I
    IADD
    IRETURN
    MAXSTACK = 2
    MAXLOCALS = 1
```

it should be something like: `31 * name.hashCode() + Integer.hashCode(age)`

Under the hood `Objects.hash()` calls `Arrays.hashCode` which is implemented with the same algorithm. So basically both function calls should be noncompliant.

leveretka: You can find out more information about the default implementations here: https://www.baeldung.com/kotlin/data-class-equals-method#2-understanding-data-classs-equals-method-and-solving-the-problem

leveretka (changes requested):
There are a few missing cases, let's add them.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/RedundantMethodsInDataClassesCheck.kt
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+        if (callExpressions.size > 1 || callExpressions.isEmpty()) return false
+
+        if ((callExpressions[0].calleeExpression as KtNameReferenceExpression).getReferencedName() == "hash") {
+            if (callExpressions[0].valueArguments.size != klassParameters.size) return false
+            callExpressions[0].valueArguments.forEach {
+                findParameter(it.getArgumentExpression(), klassParameters) ?: return false
+            }
+        } else {
```
leveretka, line 103: if you need different logic for `Objects.hash()` and `Arrays.hashCode()` maybe it's better to match them separately

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/RedundantMethodsInDataClassesCheck.kt
```diff
@@ -0,0 +1,186 @@
...
+            val arguments = callExpressions[0].valueArguments[0].collectDescendantsOfType<KtValueArgumentList>()
```
leveretka, line 104: I think it would be safer to actually get the `arrayOf(...)` by using `callExpressions[0].valueArguments[0].getArgumentExpression()` and check that it's an `arrayOf(...)` with proper arguments. This logic might give some FPs if accidentally a different function was used

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/RedundantMethodsInDataClassesCheck.kt
```diff
@@ -0,0 +1,186 @@
...
+            if (returnExpressions.size > 1) return false
+            checkEqualsExpression(returnExpressions[0], klassParameters, methodParameters, className)
+        } else {
+            val expression = findDescendantOfType<KtBinaryExpression>() ?: return false
+            checkEqualsExpression(expression, klassParameters, methodParameters, className)
+        }
+    }
+
+    private fun checkEqualsExpression(expression: KtExpression, klassParameters: List<KtParameter>, methodParameters: MutableList<KtParameter>, className: String?): Boolean {
+        val notEqualsExpressions = expression.collectDescendantsOfType<KtBinaryExpression> { it.operationToken == KtTokens.EXCLEQ }
```
leveretka, line 131: I tried to think of a better implementation, as I don't really like, when we use `.collectDescendantsOfType` as it ignores other expressions and we might face FPs here.

How about:
1. Check if expression is binary expression
2. recursively visit expression.right and expression.left
3. if one of them is not a KtBinaryExpression or a leaf, we return 

Knowing that the parser is left-recursive, leaves should be on the right

This way we can eliminate cases when there is something else between binary Expressions

leveretka (changes requested):
Looks better now, I just have concern about the equals check, maybe you can update it a little bit

leveretka (approved):
LGTM

PR #317, opened 2023-04-25
Marco/sonarkt 297

kaufco:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,234 @@
...
+import org.sonarsource.kotlin.api.predictRuntimeIntValue
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+private const val packageKotlinCollection = "kotlin.collections"
+private const val packageKotlinText = "kotlin.text"
+
+private const val interfaceKotlinCollection = "$packageKotlinCollection.Collection"
+private const val interfaceKotlinMap = "$packageKotlinCollection.Map"
+
+private val emptyArguments = listOf(emptyList<ArgumentMatcher>())
```
leveretka, line 59: Usually, we're using Upper snake case for top file final fields. We can maybe revisit it, but let's be consistent for now

kotlin-checks-test-sources/src/main/kotlin/checks/SimplifySizeExpressionCheckSample.kt
```diff
@@ -0,0 +1,236 @@
...
+        if (0 == list2.size || list == null) foo() // Noncompliant {{Replace collection size check with "isEmpty()"}}
+//          ^^^^^^^^^^^^^^^
+        if (list2.size != 0 || list == null) foo() // Noncompliant {{Replace collection size check with "isNotEmpty()"}}
+
+        if (list != null && list.isNotEmpty()) foo() // Noncompliant {{Replace null check and collection size check with "!isNullOrEmpty()"}}
+        if (list != null && list.isEmpty()) foo() // Compliant
+        if (list == null && list2.isNotEmpty()) foo() // Compliant
+        if (list == null && list2.isEmpty()) foo() // Compliant
+
+        if (list == null || list.isEmpty()) foo() // Noncompliant {{Replace null check and collection size check with "isNullOrEmpty()"}}
```
leveretka, line 148: have you checked `and` and `or` in this examples

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FieldMatcher.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
+
+fun interface FieldMatcher {
+    fun matches(node: KtNameReferenceExpression, bindingContext: BindingContext): Boolean
+}
+
+class FieldMatcherBuilderContext {
```
leveretka, line 32: I don't think you need KDocs here, names seemed to be pretty self-explanatory.
And we also don't have them for the FunMatcher.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,234 @@
...
+    arguments = emptyArguments
+)
+
+private val isNotEmptyMatcher = FunMatcher(
+    definingSupertypes = setOf(packageKotlinCollection, packageKotlinText),
+    name = "isNotEmpty",
+    arguments = emptyArguments
+)
+
+private val sizeFieldMatcher = fieldMatcher {
```
leveretka, line 79: I'm not sure I like this syntax, I'd rather make it 2 separate matchers,

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,234 @@
...
+    fieldMatcher {
+        names = setOf("size")
+        definingSupertypes = setOf(interfaceKotlinCollection, interfaceKotlinMap)
```
leveretka, line 82: I think for simple cases you can just use arguments

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FieldMatcher.kt
```diff
@@ -0,0 +1,114 @@
...
+    var definingSupertypes: Set<String> = emptySet()
+
+    internal val fieldMatchers = mutableSetOf<FieldMatcher>()
+
+    fun fieldMatcher(initializer: FieldMatcherBuilderContext.() -> Unit) {
+        fieldMatchers += createFieldMatcher(initializer)
+    }
+}
+
+fun fieldMatcher(initializer: FieldMatcherBuilderContext.() -> Unit) =
```
leveretka, line 50: If this is a way to create a `FieldMatcher`, that's better to start with a Capital letter. So you `pretend` to be a constructor

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -89,19 +94,13 @@ class FunMatcherImpl(
             checkCallParameters(functionDescriptor)
 
     private fun checkTypeOrSupertype(functionDescriptor: CallableDescriptor) =
-        qualifier.isNullOrEmpty() && definingSupertype.isNullOrEmpty() ||
```
leveretka, line 92: Also, if you changed the FunMatcher, you should add an appropriate test for it

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,234 @@
...
+        val bindingContext = kotlinFileContext.bindingContext
+        if (((operationTokenId != LT_Id) && isIntZeroLiteral(sizeTest.right, bindingContext)) ||
+            ((operationTokenId != GT_Id) && isIntZeroLiteral(sizeTest.left, bindingContext))
+        ) {
+            checkNullTestOrReportSize(sizeTest, isSizeEquals, true, expression.parent as? KtDotQualifiedExpression, kotlinFileContext)
+        }
+    }
+
+    private fun checkIsEmptyTest(expression: KtCallExpression, isIsEmpty: Boolean, kotlinFileContext: KotlinFileContext) {
+        var isEmptyTest: PsiElement = expandParenthesis(expandOptionalQualifier(expression.parent)) as? KtExpression ?: return
```
leveretka, line 140: should the inferred type be KtCallExpression?

leveretka (changes requested):
Nice idea with FieldMatcher, I think we just need to simplify it a bit

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,234 @@
...
+)
+
+private val sizeFieldMatcher = fieldMatcher {
+    fieldMatcher {
+        names = setOf("size")
+        definingSupertypes = setOf(interfaceKotlinCollection, interfaceKotlinMap)
+    }
+    fieldMatcher {
+        names = setOf("length")
+        definingSupertypes = setOf("kotlin.String")
```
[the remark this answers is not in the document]
kaufco: You're right. The code is correct, the naming is misleading though. In short: it includes the supertypes and the type itself. What it does is: 

_Match on a reference expression, if the field is declared in `X` and `X` is the type of the receiver OR a supertype of the receiver._

And there is just no distinction in the `FieldMatcher` between "receiver is exactly `X`" and "receiver is assignment compatible to `X`", even though the latter would be sufficient in the case of String.

**Suggestion:** rename `definingSuperTypes` into `definingTypes`, and we implicitly assume that the receiver just needs to be assignment compatible to the type and not _exactly_ that type. FIXED.

**Two remarks:**

1. There is a field `definingSuperType` also in the function matcher. Should it be renamed for the sake of consistency, too, or does it really only include supertypes.
2. Fun fact: `DeclarationDescriptor.getAllSuperClassifiers()` from the JetBrains `descriptorUtil` API also does include the type itself, not only true supertypes. However, the common definition of "the set of superclasses" in OOP does not include the class itself.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -42,21 +42,26 @@ private const val VARARG_PREFIX = "vararg ";
 class FunMatcherImpl(
     // In case of Top Level function there is no type there,
     // so you just need to specify the package
-    val qualifier: String? = null,
+//val qualifierx: String? = null,
+    val qualifiers: Set<String> = emptySet(),
```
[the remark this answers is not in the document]
kaufco: **As for the FunMatcher arguments,** I think you are right. I even think we should remove them as arguments entirely and rely on the typesafe builder. The only argument to keep them is to avoid refactoring. Reasons:

1. We do now have two different approaches to how we can achieve the same thing with the API: using parameters or using the typesafe builder. Good API design should try to avoid that, as to avoid disambiguity in how the user should use the API, and also, to avoid redundancy.
2. Avoid boilerplate code in the function matcher. We have defined the whole set of parameters (`qualifier`, `name`, `arguments`, ...) in three different locations: `FunMatcherImpl`, `FunMatcherBuilderContext` and `FunMatcher` functions.

In the `FieldMatcher`, I skipped such arguments entirely and you only have instance fields (declared once) that you can access from within the typesafe builder.

**As for the matching logic:** I do not see any chance how to simplify that. There is no method `withQualifiers` at the moment. If we want to introduce one, we still need a `qualifiers: set<String>` and a change in the matching logic. The only difference is that `qualifiers` is not a constructor argument anymore.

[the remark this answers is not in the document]
kaufco: P.S.: since all these properties are copied from the `FunMatcherBuilderContext` to the `FunMatcherBuilderImpl`, it would not make sense to remove `qualifiers` from the `FunMatcherImpl` constructor. We could move it only from the `FunMatcherBuilderContext` constructor only and make it a regular instance field there, and remove it from the `FunMatcher` function as an argument. 

However, `qualifiers` would be the only FunMatcher property that is treated differently then, and I don't know if that makes sense. 

My suggestion would therefore be: no special treat for `qualifiers` (i.e., not fixed), but maybe we should refactor the `FunMatcher` later to enforce the use of the typesafe builder. Wdyt?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -42,21 +42,26 @@ private const val VARARG_PREFIX = "vararg ";
...
     val names: Set<String> = emptySet(),
     // nameRegex is used when function name can be matched with regex,
     // and can be used together with names argument
     val nameRegex: Regex? = null,
     private val maxArgumentCount: Int = Int.MAX_VALUE,
     val arguments: List<List<ArgumentMatcher>> = emptyList(),
-    val definingSupertype: String? = null,
+//val definingSupertypex: String? = null,
+    val definingSupertypes: Set<String> = emptySet(),
```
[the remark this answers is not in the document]
kaufco: Same as `qualifiers`. My suggestion would be: no special treatment for definingSupertypes (i.e., not fixed), but maybe we should refactor the FunMatcher later to enforce the use of the typesafe builder.

Reasoning see above.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -42,21 +42,26 @@ private const val VARARG_PREFIX = "vararg ";
...
+//val qualifierx: String? = null,
+    val qualifiers: Set<String> = emptySet(),
     val names: Set<String> = emptySet(),
     // nameRegex is used when function name can be matched with regex,
     // and can be used together with names argument
     val nameRegex: Regex? = null,
     private val maxArgumentCount: Int = Int.MAX_VALUE,
     val arguments: List<List<ArgumentMatcher>> = emptyList(),
-    val definingSupertype: String? = null,
+//val definingSupertypex: String? = null,
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -42,21 +42,26 @@ private const val VARARG_PREFIX = "vararg ";
 class FunMatcherImpl(
     // In case of Top Level function there is no type there,
     // so you just need to specify the package
-    val qualifier: String? = null,
+//val qualifierx: String? = null,
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FieldMatcher.kt
```diff
@@ -0,0 +1,114 @@
...
+ */
+package org.sonarsource.kotlin.api
+
+import org.jetbrains.kotlin.descriptors.ClassifierDescriptor
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
+
+fun interface FieldMatcher {
```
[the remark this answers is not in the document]
kaufco: Fixed. Removed the additional code. 

The purpose of the code (case distinction in the builder function + two additional `gen*` methods) was to create specialized FieldMatchers depending on the matcher arguments. I don't have the specific use case for them in my rule though,  and I also did not do performance measurements to justify the additional code.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,234 @@
...
+
+private fun expandParenthesis(element: PsiElement?): PsiElement? = iterateParenthesis(element) {
+    parent
+}
+
+private fun reduceParenthesis(element: PsiElement?): PsiElement? = iterateParenthesis(element) {
+    expression
+}
+
+private inline fun iterateParenthesis(element: PsiElement?, next: KtParenthesizedExpression.() -> PsiElement?): PsiElement? {
```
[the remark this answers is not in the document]
kaufco: D'oh! Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
kaufco: Fixed. This was a relic from an earlier stage of the code. PsiElement is sufficient.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FieldMatcher.kt
kaufco: I would vote for KDocs here because I think whether an empty set means "match all" or "match none"  is not self-explanatory. If we apply pure set logic, it would be "match none". Since we differ from that, the semantics should be documented.

kotlin-checks-test-sources/src/main/kotlin/checks/SimplifySizeExpressionCheckSample.kt
kaufco: Do you mean the bitwise operators? But they operate on `Int` and `Long` only in Kotlin, not on `Boolean`.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FieldMatcher.kt
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -42,21 +42,26 @@ private const val VARARG_PREFIX = "vararg ";
...
+//val definingSupertypex: String? = null,
+    val definingSupertypes: Set<String> = emptySet(),
     private val matchConstructor: Boolean = false,
     val isDynamic: Boolean? = null,
     val isExtensionFunction: Boolean? = null,
     val isSuspending: Boolean? = null,
     val isOperator: Boolean? = null,
     val returnType: String? = null,
 ) {
+    private val qualifiersOrDefiningSupertype: Set<String> =
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
kaufco: Fixed, but only for `private const` (we can talk about this if you think we should do it for all).

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
```diff
@@ -89,19 +94,13 @@ class FunMatcherImpl(
...
-
-    private fun checkType(functionDescriptor: CallableDescriptor): Boolean =
-        (qualifier ?: definingSupertype)?.let {
-            when (functionDescriptor) {
-                is ConstructorDescriptor ->
-                    functionDescriptor.constructedClass.fqNameSafe.asString() == it
-                else ->
-                    functionDescriptor.fqNameSafe.asString().substringBeforeLast(".") == it
+        qualifiersOrDefiningSupertype.isEmpty() || qualifiersOrDefiningSupertype.contains(
+            if (functionDescriptor is ConstructorDescriptor) {
```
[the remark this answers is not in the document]
kaufco: Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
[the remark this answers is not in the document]
kaufco: As discussed: added an additional `qualifiers` property for exact match, like in `FunMatcher`. Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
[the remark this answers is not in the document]
kaufco: As discussed: removed lists and sets from the parameter list and refactored the code to use typesafe builder. 

I did this even for `names` and `arguments` and refactored the respective checks. 

Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FunMatcher.kt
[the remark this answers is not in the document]
kaufco: Fixed. (see above)

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/api/FieldMatcher.kt
kaufco: Refactored the `FieldMatcher`. These fields are no more public and the user is forced to rely on the `with*` functions only. Because they are self-explanatory, the problem does no more exist. I removed the comments. Fixed.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/SimplifySizeExpressionCheck.kt
```diff
@@ -0,0 +1,217 @@
...
+            EQEQ_Id -> true
+            EXCLEQ_Id -> false
+            else -> return false
+        }
+        if (isEqualsNull != useIsEmpty) return false
+
+        if (getNullTestReference(nullTest)?.getReferencedName() != sizeTestReceiver.getReferencedName()) return false
+
+        val replaceFunction = if (useIsEmpty) "isNullOrEmpty()" else "!isNullOrEmpty()"
+        kotlinFileContext.reportIssue(connectorExpression, "Replace null check and collection size check with \"$replaceFunction\"")
```
leveretka, line 196: Let's use code highlighting here

leveretka (approved):
Overall looks good, a few style issues.
I will take care of updating and merging it.

PR #318, opened 2023-04-26
[NO JIRA] Add guard around usage of new message reporting API…

johann-beleites-sonarsource:
… endpoints to ensure we don't crash in older product versions.


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/InputFileContext.kt
```diff
@@ -26,13 +26,17 @@ import org.sonar.api.batch.sensor.SensorContext
...
+import org.sonar.api.utils.log.Loggers
 import org.sonarsource.kotlin.api.InputFileContext
 import org.sonarsource.kotlin.api.Message
 import org.sonarsource.kotlin.api.SecondaryLocation
 import org.sonarsource.kotlin.converter.KotlinTextRanges.contains
 
 private const val PARSING_ERROR_RULE_KEY = "ParsingError"
 
+private val LOG = Loggers.get(InputFileContextImpl::class.java)
+private var alreadyLoggedIssueMessageWithHighlightingFailed = false
```
leveretka, line 38: I'm not a big fan of 'static' variables that are not constants. But can't think of a way to do it better and since it's a simple boolean might not be that tricky.

However, could you please add a comment on why it's like this to discourage the usage of "static" variables in general

leveretka (approved):
LGTM

PR #320, opened 2023-05-03
SONARKT-314 Fix FP in S6517 when an interface has a parent

leveretka:


irina-batinic-sonarsource (commented):
I have one question for a case:
```
interface Child : Parent {
        fun doSomething(args: String): String
}
interface Parent {}
```
Should this be Noncompliant?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/InterfaceCouldBeFunctionalCheck.kt
```diff
@@ -58,6 +58,8 @@ class InterfaceCouldBeFunctionalCheck : AbstractCheck() {
 }
 
 private fun hasExactlyOneFunctionAndNoProperties(klass: KtClass): Boolean {
+    if (klass.superTypeListEntries.isNotEmpty()) return false
+
     // Note: other possible declarations are KtClass (classes, interfaces) and
     //       KtObjectDeclaration (companion objects), but they are allowed inside function interfaces
     var functionCount = 0
```
ADarko22, line 65: Optional: it's not related to the current change but, in my opinion, this function could be more readable and consistent with its name by changing the return to:
```
    return klass.declarations.filterIsInstance<KtNamedFunction>().size == 1 && klass.declarations.all { it !is KtProperty ```

leveretka:
> I have one question for a case:
> 
> ```
> interface Child : Parent {
>         fun doSomething(args: String): String
> }
> interface Parent {}
> ```
> 
> Should this be Noncompliant?

That's a good question. In general, you can put there `fun` and that will be okay. However, that's questionable whether this interface was really meant to be functional or no. IMO, there is no guarantee that eventually `Parent` will have more abstract methods. So for this implementation, I decided to go 'safe' and cut all the cases, where we have parents.

Also, exploring all super interfaces on the subject of having a single abstract method might be costly in terms of performance and might not bring many benefits in the end.

However, if you feel we should cover this case, I am open to discussing the point as I wasn't completely sure and would like to hear your opinion.

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/InterfaceCouldBeFunctionalCheck.kt
leveretka: Good point!

irina-batinic-sonarsource:
> 

I understand your point. In this case, from my side, PR can be merged.

irina-batinic-sonarsource (approved):
LGTM!

PR #324, opened 2023-05-04
SONARKT-298 Add rule S6611 - MapValuesShouldBeAccessedSafelyCheck

irina-batinic-sonarsource:


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MapValuesShouldBeAccessedSafelyCheck.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.sonarsource.kotlin.api.FunMatcher
+import org.sonarsource.kotlin.plugin.KotlinFileContext
+
+@Rule(key = "S6611")
+class MapValuesShouldBeAccessedSafelyCheck : CallAbstractCheck() {
+
+    override val functionsToVisit = listOf(
+            FunMatcher {
+                withQualifiers("kotlin.collections.Map", "kotlin.collections.MutableMap")
+                withNames("get", "getValue", "getOrElse", "getOrDefault", "null")
```
leveretka, line 41: I think there was a misunderstanding from the rule description.

The rule should raise an issue only on case of `map.get(key)!!` or `map[key]!!`
All other functions are considered valid as they handle nullable type. So, in this matcher you should only visit `get`. So you use `FunMatcher(name = "get")`

Also, in your matcher you're matching a direct type, however I can extend Map and have my own class, we should still raise an issue in this case, so you need to use `definingSupertypes` instead of `qualifiers`

Feel free to update the rule descrition, if you feel it's not clear enough: https://github.com/SonarSource/rspec/pull/1799/files

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MapValuesShouldBeAccessedSafelyCheck.kt
```diff
@@ -0,0 +1,65 @@
...
+    override fun visitClass(klass: KtClass, context: KotlinFileContext) {
+        val arrayAccessExpressions = klass.collectDescendantsOfType<KtArrayAccessExpression> {
+            it.arrayExpression.kotlinType(context.bindingContext)?.getKotlinTypeFqName(false) == "kotlin.collections.Map"
+                    || it.arrayExpression.kotlinType(context.bindingContext)?.getKotlinTypeFqName(false) == "kotlin.collections.MutableMap"
+        }
+
+        arrayAccessExpressions.forEach {
+            val sibling = it.getNextSiblingIgnoringWhitespace()
+            if (sibling is KtOperationReferenceExpression && sibling.operationSignTokenType == KtTokens.EXCLEXCL)
+                context.reportIssue(it.parent, "\"Map\" values should be accessed safely. Using the non-null assertion operator here can throw an unexpected NullPointerException.")
```
leveretka, line 61: as `Map` is code here, we can use code highlighting in messages. 
message {
+"some text"
code("Map")
+"some text."
}

To test it just use \`Map\` in test files.

About the message itself, looks fine, I'd remove the word `unexpected`, as it's pretty much expected and is written in Kotlin documentation :)

ADarko22 (commented):
Good job! Overall it looks very nice :) just 2 comments

leveretka (approved):

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MapValuesShouldBeAccessedSafelyCheck.kt
```diff
@@ -0,0 +1,83 @@
...
+        arrayAccessExpressions.forEach {
+            val sibling = it.getNextSiblingIgnoringWhitespace()
+            if (sibling is KtOperationReferenceExpression && sibling.operationSignTokenType == KtTokens.EXCLEXCL)
+                context.reportIssue(it.parent, issueMessage)
+        }
+    }
+
+    private fun checkSuperType(arrayAccessExpression: KtArrayAccessExpression, bindingContext: BindingContext): Boolean {
+        val type = arrayAccessExpression.arrayExpression.determineType(bindingContext) ?: return false
+        if (checkIfSubtype(type)) return true
```
ADarko22, line 74: you can simplify with `return checkIfSubtype(type) || type.supertypes().any { checkIfSubtype(it) }`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MapValuesShouldBeAccessedSafelyCheck.kt
```diff
@@ -0,0 +1,83 @@
...
+        return type.supertypes().any {
+            checkIfSubtype(it)
+        }
+    }
+
+    private fun checkIfSubtype(type: KotlinType) = type.getKotlinTypeFqName(false) == "kotlin.collections.Map"
```
ADarko22, line 80: nice move to extract the logic in this method and we could improve readability even more with a better name; what do you think of `isMapImplementation`, `implementsMap`, or `isMapSubtype`?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/checks/MapValuesShouldBeAccessedSafelyCheck.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        arrayAccessExpressions.forEach {
+            val sibling = it.getNextSiblingIgnoringWhitespace()
+            if (sibling is KtOperationReferenceExpression && sibling.operationSignTokenType == KtTokens.EXCLEXCL)
+                context.reportIssue(it.parent, issueMessage)
+        }
+    }
+
+    private fun checkSuperType(arrayAccessExpression: KtArrayAccessExpression, bindingContext: BindingContext): Boolean {
```
ADarko22, line 72: Similar to the comment below, we could improve the method's name, some suggestions may be: `isMapImplementation`, `implementsMap`, or `isMapSubtype`

PR #333, opened 2023-06-05
SONARKT-334 Fix calculation of lines of code and multiline comment

irina-batinic-sonarsource:


ADarko22 (approved):

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/MetricVisitor.kt
```diff
@@ -127,6 +129,8 @@ private class KtMetricVisitor : KtTreeVisitorVoid() {
 
         // We don't want to count file headers as comment.
         file.children.dropWhile { it is PsiComment || it is PsiWhiteSpace }.forEach { it.accept(this) }
+
+        file.containingFile.collectDescendantsOfType<KDoc>().forEach { addCommentMetrics(it, commentLines, nosonarLines) }
```
johann-beleites-sonarsource, line 133: I was confused when I read this line at first, as this is such a different way of collecting the metrics vs the visitor pattern we have for the rest. After getting back into it, it makes sense, as we don't have an adequate `visit` method available in the visitor. How about adding a comment to this line explaining why we do it this way?

johann-beleites-sonarsource, line 133: With this line, I think we can delete line 141:
```
        function.docComment?.let { addCommentMetrics(it, commentLines, nosonarLines) }
```

irina-batinic-sonarsource: I agree. It is done.

johann-beleites-sonarsource (approved):

PR #342, opened 2023-07-04
SONARKT-344 - setupGradleRuleStubs + fix setupGradleRuleStubs paths

kaufco:


buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -177,3 +193,37 @@ private object Templates {
...
+    override val checkListFile =
+        Path.of(KOTLIN_GRADLE_BASEDIR, "src", "main", "java", "org", "sonarsource", "kotlin", "gradle", "KotlinGradleCheckList.kt")
+
+    override fun generateCheckClass(ruleKey: String, checkClassName: String, messageLine: String?) = LICENSE_HEADER + """
+        package org.sonarsource.kotlin.gradle.checks
+        
+        import org.sonar.check.Rule
+        import org.sonarsource.kotlin.api.checks.AbstractCheck
+        
+        ${messageLine?.let { "        \n        $it\n" } ?: ""}
```
johann-beleites-sonarsource: Remove the empty line to avoid a superfluous empty line
```suggestion
        ${messageLine?.let { "        \n        $it\n" } ?: ""}
```

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
johann-beleites-sonarsource, line 1: Some indentation / formatting is not correct in this file. (In IntelliJ, `Ctrl`+`Alt`+`L` is your friend).

johann-beleites-sonarsource, line 1: This file is getting somewhat large. I would suggest splitting it up into multiple files. It would also be easier to create a new task for the Gradle stubs instead of having to pass a flag indicating that we want to initialize a Gradle check. The task could be called `setupGradleRuleStubs` for instance. Then you can have one file per task, including the specific template implementations with some common logic in a third file.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -22,6 +16,8 @@ import kotlin.io.path.readText
 import kotlin.io.path.writeLines
 import kotlin.io.path.writeText
 
+private const val KOTLIN_GRADLE_BASEDIR = "sonar-kotlin-gradle"
```
johann-beleites-sonarsource, line 19: This is not used everywhere where it could. It is also specific to the Gradle stubs only, it should be moved in the Gradle-specific file along with the Gradle-specific logic (see file-level comment above).

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -35,6 +31,10 @@ abstract class CreateRuleStubsTask : DefaultTask() {
     @get:Optional
     val message: String? by project
 
+    @get:Input
+    @get:Optional
+    val gradleRule: String? by project
```
johann-beleites-sonarsource, line 36: Replace with separate task as explained in file-level comment above. Don't hesitate to reach out if you have questions about how to do this.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -84,7 +89,7 @@ abstract class CreateRuleStubsTask : DefaultTask() {
             false
         }
 
-    private fun addRuleToChecksListFile(checkClassName: String): Boolean {
+    private fun Templates.addRuleToChecksListFile(checkClassName: String): Boolean {
```
johann-beleites-sonarsource, line 92: This function contains a hard-coded checks package. This is incorrect for Gradle checks and will cause a compile-time error in the `KotlinGradleCheckList.kt`.

johann-beleites-sonarsource (changes requested):
Don't hesitate to reach out if you have questions

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -135,21 +140,36 @@ abstract class CreateRuleStubsTask : DefaultTask() {
...
+    override val checksPackage = "org.sonarsource.kotlin.checks"
+    override val sampleFileExt = "kt"
 
-    fun generateCheckClass(ruleKey: String, checkClassName: String, messageLine: String?) = LICENSE_HEADER + """
-        package org.sonarsource.kotlin.checks
+    override fun generateCheckClass(ruleKey: String, checkClassName: String, messageLine: String?) = LICENSE_HEADER + """
+        package $checksPackage
 
         import org.sonar.check.Rule
         import org.sonarsource.kotlin.api.AbstractCheck
```
johann-beleites-sonarsource, line 175: I believe this is also not the correct import anymore.

johann-beleites-sonarsource:
I think this branch needs to be rebased with master - there seem to be some unrelated commits still

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
kaufco: Done.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
kaufco: Done.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
kaufco: Done.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
kaufco: Done. Also noticed this when creating rule S6626.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
kaufco: Done.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateKotlinGradleRuleStubsTask.kt
```diff
@@ -0,0 +1,42 @@
...
+            }
+            
+        """.trimIndent()
+
+        override fun generateTestClass(testClassName: String, checkClassName: String) = LICENSE_HEADER + """
+            package $checksPackage
+            
+            internal class $testClassName : CheckTest($checkClassName())
+                    
+        """.trimIndent()
```
johann-beleites-sonarsource, line 35: These two methods seem to be the same in both template implementations. Let's move them up!

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -48,47 +42,46 @@ abstract class CreateRuleStubsTask : DefaultTask() {
...
+        targetFile.createFile()
+        targetFile.writeText(content)
+        true
+    } else {
+        logger.warn("WARNING: File '$targetFile' exists. Not creating a new one.")
+        false
+    }
 
-    private fun addRuleToChecksListFile(checkClassName: String): Boolean {
+    private fun RuleStubTemplates.addRuleToChecksListFile(checkClassName: String): Boolean {
```
johann-beleites-sonarsource, line 80: There are a lot of functions that have become extension functions. While extension functions can be a valuable tool, in this case we can simply move the functions into the corresponding type, making the code easier to understand. Perhaps then we should also rethink the "Template" naming and make it a "Factory"/"Builder"/"Creator"/"Generator" or so instead.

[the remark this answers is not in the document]
kaufco: This is because the list is empty. Current implementation requires a non-empty list because it scans for the pattern "Check::class.java" somewhere in the file. I left it as it was because this is only an issue for the very first rule added.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateKotlinGradleRuleStubsTask.kt
kaufco: Fixed: moved them as common code (i.e., not part of the template) into CreateRuleStubsTask. They still could be moved back to the templates once they actually should start to differ.

buildSrc/src/main/kotlin/org/sonarsource/kotlin/buildsrc/tasks/CreateRuleStubsTask.kt
```diff
@@ -48,47 +42,66 @@ abstract class CreateRuleStubsTask : DefaultTask() {
...
+        """.trimIndent()
+
+    private fun generateTestClass(testClassName: String, checkClassName: String) = LICENSE_HEADER + """
+            package ${templates.checksPackage}
+    
+            internal class $testClassName : CheckTest($checkClassName())
+
+        """.trimIndent()
+
+    private fun RuleStubTemplates.createSampleFile(checkClassName: String) =
```
johann-beleites-sonarsource, line 88: Still one fun that is an extension fun - I assume this was simply forgotten?

johann-beleites-sonarsource (approved):
One small comment, then LGTM

PR #350, opened 2023-07-12
SONARKT-328 - Implement rule S6625: "rootProject.name" should always be present in Gradle settings

kaufco:


sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/checks/RootProjectNamePresentCheck.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    override fun visitKtFile(file: KtFile, kotlinFileContext: KotlinFileContext) {
+        if (!file.name.endsWith(SETTINGS_FILE_NAME)) return
+
+        val visitor = ScanRootProjectNameAssign()
+        file.acceptChildren(visitor)
+        if (!visitor.isRootProjectNameAssigned) {
+
+            kotlinFileContext.reportIssue(
+                kotlinFileContext.textRange(file.startOffset, file.startOffset + 1),
```
leveretka, line 50: I've just realised that if file is empty this might fail. Maybe let's report on a file? you can pass `null` as a text range. Look at `FileHeaderCheck`

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/checks/RootProjectNamePresentCheck.kt
```diff
@@ -0,0 +1,76 @@
...
+                    +" in "
+                    code(SETTINGS_FILE_NAME)
+                }
+            )
+        }
+    }
+
+    private class ScanRootProjectNameAssign : KtTreeVisitorVoid() {
+
+        var isRootProjectNameAssigned = false
```
leveretka, line 63: I think people can also do `rootProject.setName("kotlin") ` and while it's not the way to go, we shouldn't report the issue here

leveretka (changes requested):
Good job, I think it requires just mentioned 2 points

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/checks/RootProjectNamePresentCheck.kt
kaufco: Cool :-) Fixed.

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/checks/RootProjectNamePresentCheck.kt
```diff
@@ -0,0 +1,90 @@
...
+        if (!visitor.isRootProjectNameAssigned) {
+
+            kotlinFileContext.reportIssue(
+                null,
+                message {
+                    +"Assign "
+                    code("$IDENTIFIER_ROOT_PROJECT.$IDENTIFIER_NAME")
+                    +" in "
+                    code(SETTINGS_FILE_NAME)
+                }
```
leveretka, line 58: missing "." at the end of the issue message

leveretka (approved):
One missing dot. Otherwise, Looks good!

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/checks/RootProjectNamePresentCheck.kt
kaufco: Fixed.

PR #361, opened 2023-07-26
fix ci-build failure due to OOMKilled

ADarko22:


johann-beleites-sonarsource:
Closing, as we are using 12G for now (see #389).

PR #374, opened 2023-10-31
SONARKT-368 Stop raising issues when toString is not implemented for S6218

erwan-serandour-sonarsource:


leonardo-pilastri-sonarsource (approved):
LGTM!

PR #375, opened 2023-11-01
SONARKT-379 Fix false positive in S6218

fellmann:
Overriding `equals` and `hashCode` is not necessary for data classes that have array fields in their body, because only arguments of the primary constructor are included in the generated `equals` and `hashCode` for data classes.

This should not be trigger the rule:
```kotlin
data class ArrayInBody(val age: Int) { // Compliant
    val employers = arrayOf("SonarSource")
}
```


johann-beleites-sonarsource:
Hello @fellmann, thanks for your contribution! I've created [a ticket](https://sonarsource.atlassian.net/jira/software/c/projects/SONARKT/issues/SONARKT-379) to track this issue.

PR #380, opened 2023-11-08
SONARKT-303: Implement rule S6527: Function chain using "filter" should be simplified

ADarko22:


johann-beleites-sonarsource (commented):
Looks nice! Just a minor suggestions.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/SimplifyFilteringBeforeTerminalOperationCheck.kt
```diff
@@ -0,0 +1,68 @@
...
+        FunMatcher(qualifier = KOTLIN_COLLECTIONS_QUALIFIER, name = "singleOrNull") { withNoArguments() }
+    )
+
+    override fun visitFunctionCall(callExpression: KtCallExpression, resolvedCall: ResolvedCall<*>, kotlinFileContext: KotlinFileContext) {
+        val chainedCallBeforeTerminalOperationCall = callExpression.parent
+            .let { it as? KtDotQualifiedExpression }
+            ?.receiverExpression
+            .let { it as? KtDotQualifiedExpression }
+            ?.selectorExpression
+            .let { it as? KtCallExpression }
```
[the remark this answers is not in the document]
ADarko22: Thank you for all the details and the hints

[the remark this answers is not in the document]
ADarko22: Looks like that there is not even need for the `skiParenteses()`by using your suggestion:
```
val chainedCallBeforeTerminalOperationCall = callExpression.parent
            .let { it as? KtDotQualifiedExpression }
            ?.receiverExpression
            ?.getCall(kotlinFileContext.bindingContext)
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/SimplifyFilteringBeforeTerminalOperationCheck.kt
```diff
@@ -0,0 +1,64 @@
...
+            withNoArguments()
+        }
+    )
+
+    @OptIn(IDEAPluginsCompatibilityAPI::class)
+    override fun visitFunctionCall(callExpression: KtCallExpression, resolvedCall: ResolvedCall<*>, kotlinFileContext: KotlinFileContext) {
+        val chainedCallBeforeTerminalOperationCall = callExpression.parent
+            .let { it as? KtDotQualifiedExpression }
+            ?.receiverExpression
+            .let { it?.getCall(kotlinFileContext.bindingContext) }
```
ADarko22, line 49: What is the difference between `.let { it?.getCall(kotlinFileContext.bindingContext) }` and 
`?.getCall(kotlinFileContext.bindingContext)`?

johann-beleites-sonarsource, line 49: Better would be: 
```suggestion
            ?.let { it.getCall(kotlinFileContext.bindingContext) }
```
And then it becomes apparent that there is actually no need for a `let` here. You can simplify it to:
```suggestion
            ?.getCall(kotlinFileContext.bindingContext)
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/SimplifyFilteringBeforeTerminalOperationCheck.kt
```diff
@@ -0,0 +1,64 @@
...
+
+        if (chainedCallBeforeTerminalOperationCall != null
```
johann-beleites-sonarsource, line 51: You can avoid this null-check by returning early while retrieving the `chainedCallBeforeTerminalOperationCall`. Just add a `?: return` after line 49 and you will be guaranteed that you cannot arrive here with `chainedCallBeforeTerminalOperationCall` being null.

Then, if you like functional style, you don't even need that. Using the `.takeIf` function, you can chain everything and don't need an explicit `if`, without losing readability. So something like:
```
...
?.receiverExpression
?.getCall(...)
?.takeIf { FILTER_MATCHER.matches(...) }
?.let { chainedCallBeforeTerminalOperationCall ->
  // the reporting logic
}
```
It's also fine to leave a separate `if` conditional to check if it matches, though.

sonar-kotlin-checks/src/test/java/org/sonarsource/kotlin/checks/SimplifyFilteringBeforeTerminalOperationCheckTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    @Test
+    fun `with no semantics`() {
+        KotlinVerifier(check) {
+            this.fileName = "SimplifyFilteringBeforeTerminalOperationCheckNoSemanticsSample.kt"
+            this.classpath = emptyList()
+            this.deps = emptyList()
+            this.isAndroid = false
+        }.verifyNoIssue()
+    }
```
johann-beleites-sonarsource, line 35: Better yet: simply extend `CheckTestWithNoSemantics` instead of `CheckTest`. It already provides a standardized implementation of a test without semantics.

johann-beleites-sonarsource (approved):
A couple of suggestions for more idiomatic style, otherwise looks good, nice job!

PR #381, opened 2023-11-09
SONARKT-304 - Implement rule S6528: "find" should be replaced with "any", "none" or "contains"

ADarko22:


sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S6528.html
```diff
@@ -0,0 +1,68 @@
+<p>The <code>kotlin.collections</code> package offers many functions to interact with collections. In particular, functions
```
ADarko22, line 1: I updated the RSPEC as well: https://github.com/SonarSource/rspec/pull/1646

kotlin-checks-test-sources/src/main/kotlin/checks/UnsuitedFindFunctionWithNullComparisonCheckSample.kt
```diff
@@ -0,0 +1,95 @@
...
+        list.findLast { it > 5 } == null // Noncompliant  {{Replace "list.findLast { it > 5 } == null" with "list.none { it > 5 }".}}
+
+        list.find { it == 5 } != null // Noncompliant  {{Replace "list.find { it == 5 } != null" with "list.contains(5)".}}
+        list.find { 5 == it } != null // Noncompliant  {{Replace "list.find { 5 == it } != null" with "list.contains(5)".}}
+        list.find { it != 5 } == null // Noncompliant  {{Replace "list.find { it != 5 } == null" with "list.contains(5)".}}
+        list.find { 5 != it } == null // Noncompliant  {{Replace "list.find { 5 != it } == null" with "list.contains(5)".}}
+
+        list.find { it == 5 } == null // Noncompliant  {{Replace "list.find { it == 5 } == null" with "!list.contains(5)".}}
+        list.find { 5 == it } == null // Noncompliant  {{Replace "list.find { 5 == it } == null" with "!list.contains(5)".}}
+        list.find { it != 5 } != null // Noncompliant  {{Replace "list.find { it != 5 } != null" with "!list.contains(5)".}}
```
johann-beleites-sonarsource, line 19: I don't think this is quite correct. `list.find { it != 5 } != null` is checking if the list contains an element other than 5. It doesn't matter if 5 is in it, si I think `list.any { it != 5 }` would be a more accurate replacement.

johann-beleites-sonarsource (changes requested):
Some complex stuff going on, I like that you give the users concrete instructions in the message, not just a generic one. This should also help us write quickfixes, once we support them. I have a few suggestions, don't hesitate to reach out if you have any questions.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/UnsuitedFindFunctionWithNullComparisonCheck.kt
```diff
@@ -0,0 +1,118 @@
...
+        this.left!!.references(reference) || this.right!!.references(reference)
+
+    private fun KtBinaryExpression.nonParameterOperand(parameter: String): KtExpression =
+        if (!this.left!!.references(parameter)) this.left!! else this.right!!
+
+    private fun KtExpression.references(reference: String): Boolean =
+        this.takeIf { it is KtNameReferenceExpression }?.let { it as KtNameReferenceExpression }?.text == reference
+
+    private fun message(nullComparisonExpr: KtBinaryExpression, replacementExpr: String): String =
+        "Replace \"${nullComparisonExpr.text}\" with \"$replacementExpr\"."
```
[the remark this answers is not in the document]
ADarko22: Good point! I will try the code highlighting :)

kotlin-checks-test-sources/src/main/kotlin/checks/UnsuitedFindFunctionWithNullComparisonCheckSample.kt
ADarko22: You're right this will diverge in case of an empty collection:
`!list.contains(5)` returns true, while `list.find { it != 5 } != null` return false! Nice catch! :)

johann-beleites-sonarsource (approved):
LGTM! :tada:

PR #383, opened 2023-11-13
SONARKT-303 - Implement rule S6527: Function chain using "filter" should be simplified

ADarko22:


johann-beleites-sonarsource (approved):
2 minor comments, otherwise LGTM!

PR #384, opened 2023-11-14
SONARKT-302 - Implement rule S6526: abstract class should be interface

erwan-serandour-sonarsource:


johann-beleites-sonarsource (changes requested):
The logic looks good! Just a few suggestions to make it super clean :slightly_smiling_face:

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/AbstractClassShouldBeInterfaceCheck.kt
```diff
@@ -19,52 +19,39 @@
...
 
     override fun visitClass(klass: KtClass, context: KotlinFileContext) {
-        super.visitClass(klass, context)
 
-        if (!klass.isAbstract() || klass.isInterface() || klass.extendsClass(context.bindingContext)) return
+        if (!klass.isAbstract() || klass.isInterface() || klass.extendsClass()) return
 
         val allMethods = klass.collectDescendantsOfType<KtFunction>()
-        val allProperties = klass.collectDescendantsOfType<KtProperty>()
+        val allProperties: List<KtProperty> by lazy { klass.collectDescendantsOfType<KtProperty>() }
```
johann-beleites-sonarsource, line 41: That's a nice way of solving this!

johann-beleites-sonarsource (approved):
Nice, LGTM! I think for this rule we will have to do some extra peach validation and see if there are use-cases for abstract classes that we didn't think about.

PR #386, opened 2023-11-14
SONARKT-286 Implement S6619

johann-beleites-sonarsource:
Null checks should be useful


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/UselessNullCheckCheck.kt
```diff
@@ -0,0 +1,194 @@
...
+        val result = if (resolvedExpression.isNull()) {
+            nullCaseResult
+        } else if (
+        // We are not using the resolvedExpression on purpose here, as it can cause FPs. See SONARKT-373.
+            expression.isNotNullable(kfc.bindingContext)
+        ) {
+            nonNullCaseResult
+        } else {
+            return
+        }
```
ADarko22, line 168: This conditional logic is a bit hard to read at first glance, would you reconsider rearranging it? For example:
```suggestion
        val result = if (expression.predictRuntimeValueExpression(kfc.bindingContext).isNull()) {
            if (comparesToNull) "succeeds" else "fails"
        } else if (expression.isNotNullable(kfc.bindingContext)) { // Not using the resolvedExpression as it can cause FPs. See SONARKT-373.
            if (comparesToNull) "fails" else "succeeds"
        } else {
            return
        }
```

sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S6619.html
```diff
@@ -0,0 +1,68 @@
...
+</pre>
+<h4>Compliant solution</h4>
+<pre data-diff-id="3" data-diff-type="compliant">
+val s: String = "ABC"
+return s
+</pre>
+<h4>Noncompliant code example</h4>
+<pre data-diff-id="4" data-diff-type="noncompliant">
+val s: String = ""
+s!!.doSomething() //You do not need to assert that S is non-null.
```
ADarko22, line 51: Can you check the RSPEC with Grammarly? Also here the comment refer to "S" while in the code uses "s"

its/ruling/build.gradle.kts
```diff
@@ -15,6 +16,9 @@ tasks.test {
     onlyIf {
         project.hasProperty("its") || project.hasProperty("ruling")
     }
+
+    // dependsOn(tasks.withType<Copy>())
```
ADarko22, line 20: is this commented code intentional?

ADarko22 (changes requested):
Great job with this troublesome rule! 

Left a few questions about and comments to polish a bit the code.

ADarko22 (approved):

PR #390, opened 2023-11-22
SONARKT-301 - Implement rule S6524: collection should be immutable

erwan-serandour-sonarsource:


sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S6524.json
```diff
@@ -0,0 +1,15 @@
+{
+  "title": "Collection should be immutable if contents is not changed",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
```
erwan-serandour-sonarsource, line 6: add new clean code taxonomy

johann-beleites-sonarsource (changes requested):
In terms of how we go about visiting the AST, it looks fine on first glance. Although perhaps it is worth noting that we currently do not look at properties defined outside of functions (private class & file-level properties). Was that intended?

In terms of determining if a collection has been written to, so needs to be mutable, I think we need to improve the logic a bit.

The current logic assumes that a collection is not written to by default and only if one of the listed funs is called, do we recognize it as written to. If I define a custom collection extension fun or one that takes a collection as argument, which writes to the respective collection, we will miss the fact that this function is a writing function everywhere else. We can probably depend on the type information instead, to see if a collection is only used in contexts where an immutable type is required. However, if we don't understand what a call to a function does w.r.t. the collection argument/receiver, we should assume it is modified to avoid FPs. In other words, our default assumption should be "the collection IS modified" and we need to prove otherwise to raise an issue.

In general, I strongly recommend focusing more on the test cases. In fact, for rules like this, it helps to really go with TDD and think of all kinds of crazy ways you could modify collections first, even before you start writing the rule. Then start simple and see how that tricks the logic you write, accommodating the complexities bit by bit. Especially if you're not sure of the logic, write more test cases! Have some complex examples in there, where you have functions that call functions or where you pass function calls as an argument to another function call. Use `.let`, `with()`, etc. This way you have to depend way less on the PR review and ultimately it will save you some time too, because of less back-and-forth.

johann-beleites-sonarsource (changes requested):
As discussed

johann-beleites-sonarsource (approved):
As discussed today, let's double-check the results during verification later today.

PR #391, opened 2023-11-22
SONARKT-304 - simplify report message when the replacement string is …

ADarko22:
…too large; cover  use case with safe qualifiers

### DIFF
--- a/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/UnsuitedFindFunctionWithNullComparisonCheck.kt
+++ b/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/UnsuitedFindFunctionWithNullComparisonCheck.kt
@@ -23,10 +23,11 @@ import org.jetbrains.kotlin.lexer.KtTokens.EQEQ
 import org.jetbrains.kotlin.lexer.KtTokens.EXCLEQ
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtCallExpression
-import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtLambdaExpression
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtQualifiedExpression
+import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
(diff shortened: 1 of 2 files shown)


johann-beleites-sonarsource (commented):
Two suggestions, otherwise LGTM:
* I think the message length limit could be even a bit smaller. The way SQ/SC/SL display messages, a 100-char long code snippet is probably still hard to read. Users are supposed to understand a message in an instant, so here we can only really provide it for cases where it is pretty short.
* We should also probably not include samples that span multiple lines. Either filter out code snippets that contain a newline character or discard the lines before the one with the issue.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/UnsuitedFindFunctionWithNullComparisonCheck.kt
```diff
@@ -47,32 +50,30 @@ import org.sonarsource.kotlin.api.reporting.message
...
+        callExpression.findClosestAncestorOfType<KtBinaryExpression>()?.takeIf { it.right!!.isNull() || it.left!!.isNull() }
             ?.let { report(it, callExpression, kotlinFileContext) }
     }
 
     private fun report(nullComparisonExpr: KtBinaryExpression, callExpression: KtCallExpression, kotlinFileContext: KotlinFileContext) {
         // callExpression has argument a lambda expression with a single parameter, due to the functionsToVisit FunMatchers
         val lambda = callExpression.valueArguments[0].getArgumentExpression() as KtLambdaExpression
 
-        val lambdaBinaryExpr = lambda.bodyExpression!!.firstChild as? KtBinaryExpression
+        val lambdaBinaryExpr = lambda.bodyExpression!!.children[0] as? KtBinaryExpression
```
ADarko22, line 67: this way we cover the case:

```
list.find { // comment
 it == 5
} != null
```
that can be replaced by `list.contains(5)`

ADarko22, line 67: this way we can cover:
```
        list.find {// Noncompliant {{Replace with `list.contains(5)`.}}
            it == 5
        } != null
```

johann-beleites-sonarsource: So `.firstChild` and `.children[0]` behaves differently? Maybe worth a small comment in the code!

johann-beleites-sonarsource (approved):

PR #395, opened 2023-11-28
SONARKT-301 stop raising issues on abstract/overrided/open/interface methods

erwan-serandour-sonarsource:


kotlin-checks-test-sources/src/main/files/non-compiling/checks/CollectionShouldBeImmutableCheckSampleNonCompiling.kt
```diff
@@ -0,0 +1,12 @@
+package checks
+
+class CollectionShouldBeImmutableCheckSampleNonCompiling {
+
+    actual fun actualFun(list: MutableList<Int>): Unit // compliant
+    expect fun expectFun(list: MutableList<Int>): Unit // compliant
+
+    fun qualifiedStarange() {
```
johann-beleites-sonarsource, line 8:
```suggestion
    fun qualifiedStrange() {
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/CollectionShouldBeImmutableCheck.kt
```diff
@@ -52,8 +59,25 @@ class CollectionShouldBeImmutableCheck : AbstractCheck() {
             "kotlin.collections.MutableCollection"
         )
 
+    override fun visitClass(klass: KtClass, context: KotlinFileContext) {
+        if (klass.isInterface()) return
+
+        klass
+            .collectDescendantsOfType<KtNamedFunction>(::notAnInterface) { true }
+            .forEach { reportNamedFunction(it, context) }
+    }
```
johann-beleites-sonarsource, line 68: Like this, we would miss any cases that are inside of file-level functions. Like this you can get those too:

```suggestion
    override fun visitKtFile(file: KtFile, context: KotlinFileContext) {
        file
            .collectDescendantsOfType<KtNamedFunction>(::notAnInterface) { true }
            .forEach { reportNamedFunction(it, context) }
    }
```

johann-beleites-sonarsource: You can put this on file level into the sample file:
```
private fun nonCompliantParameterOnFileLevel(list: MutableList<Int>): Int { // Noncompliant
    return list.reduce { acc, it -> acc + it}
}
```

kotlin-checks-test-sources/src/main/files/non-compiling/checks/CollectionShouldBeImmutableCheckSampleNonCompiling.kt
erwan-serandour-sonarsource: :+1:

johann-beleites-sonarsource (approved):

PR #410, opened 2023-12-20
[NO JIRA] Migrate ITs to Junit 5

leveretka:


settings.gradle.kts
```diff
@@ -81,6 +81,7 @@ dependencyResolutionManagement {
             val assertj = version("assertj", "3.24.2")
             val classgraph = version("classgraph", "4.8.165")
             val junit = version("junit", "5.10.1")
+            val junitSuite = version("junitSuite", "1.10.1")
```
johann-beleites-sonarsource, line 84: I think we don't need this line anymore?

johann-beleites-sonarsource (approved):
Nice, one little unnecessary line, then LGTM!

settings.gradle.kts
leveretka: Good catch, leftover :)

PR #412, opened 2024-01-02
SONARKT-383 Update license headers

kaufco:


johann-beleites-sonarsource (changes requested):
The plugin does not seem to handle some files correctly, notably `package-info.java` files. Grep `2018-2023` to find the list of files.

johann-beleites-sonarsource (approved):

PR #414, opened 2024-02-07
BUILD-4459 fix orchestrator.configUrl lookup

julien-carsique-sonarsource:


.cirrus.yml
```diff
@@ -214,7 +214,7 @@ qa_ruling_kotlin_compiler_task:
     - source cirrus-env QA
     - ./gradlew :its:ruling:test --tests "org.sonarsource.slang.SlangRulingTest.test_kotlin_compiler" -Pruling
       "-Dsonar.runtimeVersion=${SQ_VERSION}"
-      "-Dorchestrator.configUrl=https://repox.jfrog.io/repox/orchestrator.properties/orch-h2.properties"
+      "-Dorchestrator.configUrl=file:///$CIRRUS_WORKING_DIR/orchestrator.properties"
```
ADarko22, line 217: Thanks @julien-carsique-sonarsource for finding a solution, this is great!
I'm just wondering why other projects, like [sonar-iac ](https://github.com/SonarSource/sonar-iac/blob/master/.cirrus.yml#L139) are working without problems with the orch-h2.properties URL. Do you have an idea?

orchestrator.properties
```diff
@@ -0,0 +1,3 @@
+java.io.tmpdir=${SONAR_TMP}
```
ADarko22, line 1: I will check with the team if having such a file here is fine.

.cirrus.yml
julien-carsique-sonarsource: I don't know.

I looked at the orchestrator code with @alain-kermis-sonarsource and we considered that `orchestrator.configUrl=https://repox.jfrog.io/repox/...` was not able to work with the current implementation and that it was actually not worth using that option with repox.

I don't see any download from repox in sonar-iac, only from https://repo.maven.apache.org/maven2/.
Maybe that the needed resources are already in the cache, or available from public repositories.
In sonar-iac, the settings.gradle.kts file is slightly different with:
```
pluginManagement {
    repositories {
        mavenCentral()
        gradlePluginPortal()
        maven {
            name = "artifactory"
            url = uri("https://repox.jfrog.io/repox/sonarsource")
        }
    }
}
dependencyResolutionManagement {
    repositories {
        mavenCentral()
        maven {
            url = uri("https://repox.jfrog.io/repox/sonarsource")
        }
    }
}
```
vs in sonar-kotlin:
```
pluginManagement {
    repositories {
        maven(url = "https://repox.jfrog.io/repox/plugins.gradle.org/")
        gradlePluginPortal()
    }
}
dependencyResolutionManagement {
}

allprojects {
    repositories {
        mavenLocal()
        val repository = if (project.hasProperty("qa")) "sonarsource-qa" else "sonarsource"
        maven {
            url = uri("https://repox.jfrog.io/repox/${repository}")
        }
    }
}
```
So maybe that in sonar-iac, mavenCentral is checked before looking up at repox.

orchestrator.properties
ADarko22: The solution looks good

ADarko22 (approved):

PR #421, opened 2024-04-15
[NO-JIRA] Default to Maven central when not authenticated to repox

dorian-burihabwa-sonarsource:
To enable external, and unauthenticated-to-repox, users to build sonar-kotlin locally, we add maven central as the fallback repository.


Godin:
Superseded by #455

PR #433, opened 2024-09-16
Upgrade Gradle to 8.10.1

Godin:


leveretka (approved):
Thanks for taking care!

PR #434, opened 2024-09-16
Upgrade Spotless to 6.25.0

Godin:


jonas-wielage-sonarsource:
This effort will be tracked in this ticket: https://sonarsource.atlassian.net/browse/SONARKT-741
Once implementation resumes, another PR can be openend.

PR #437, opened 2024-09-17
Remove use of io.spring.dependency-management Gradle plugin

Godin:
It was causing

* Gradle warning
   > Mutating the dependency attributes of configuration ':kotlin-checks-test-sources:implementation' after it has been resolved or consumed.

* appearance of POMs in `gradle/verification-metadata.xml`
  despite `<verify-metadata>false</verify-metadata>`


sonar-kotlin-api/build.gradle.kts
```diff
@@ -37,3 +37,10 @@ tasks.withType<Javadoc> {
         addStringOption("tag", "javax.annotation.ParametersAreNonnullByDefault:ParametersAreNonnullByDefault")
     }
 }
+
+task<JavaExec>("printAst") {
+    dependsOn(":sonar-kotlin-plugin:compileJava")
```
leveretka, line 42: As discussed in person, after moving it outside of the `utils-kotlin` module, I don't think we need this dependency; we're not even sure we needed it in the first place

Godin: Done

leveretka (approved):

PR #470, opened 2024-10-29
Fix memory leaks in tests

Godin:
Depends on
- [x] #471

### DIFF
--- a/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/frontend/KotlinCoreEnvironmentTools.kt
+++ b/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/frontend/KotlinCoreEnvironmentTools.kt
@@ -41,12 +41,18 @@ import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.lazy.declarations.FileBasedDeclarationProviderFactory
 import java.io.File
 
+/**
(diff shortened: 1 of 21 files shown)


sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/frontend/KotlinCoreEnvironmentTools.kt
```diff
@@ -41,12 +41,16 @@ import org.jetbrains.kotlin.resolve.BindingContext
 import org.jetbrains.kotlin.resolve.lazy.declarations.FileBasedDeclarationProviderFactory
 import java.io.File
 
+/**
+ * @param disposable
+ * see [documentation](https://jetbrains.org/intellij/sdk/docs/basics/disposers.html) for more details
```
leveretka, line 46:
```suggestion
 * Required to keep track of resources and ensure their cleanup. Resources will be closed once {@code dispose()} is called.
 * see [documentation](https://jetbrains.org/intellij/sdk/docs/basics/disposers.html) for more details
```

Godin: Quoting https://jetbrains.org/intellij/sdk/docs/basics/disposers.html

> Never call Disposable.dispose() directly

PR #474, opened 2024-11-01
Telemetry

Godin:


jonas-wielage-sonarsource:
This effort will be tracked in https://sonarsource.atlassian.net/browse/SONARKT-425. 
If work resumes on this effort, another PR can be created.

PR #522, opened 2025-01-22
SONARKT-429 Fix UTs on Windows with autocrlf true

antonioaversa:


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinRulesDefinitionTest.kt
```diff
@@ -38,7 +39,11 @@ internal class KotlinRulesDefinitionTest {
             .isEqualTo("Identical expressions should not be used on both sides of a binary operator")
         Assertions.assertThat(rule.type()).isEqualTo(RuleType.BUG)
         Assertions.assertThat(rule.scope()).isEqualTo(RuleScope.ALL)
-        Assertions.assertThat(rule.htmlDescription()).startsWith("<h2>Why is this an issue?</h2>\n<p>Using the same value on both sides")
+        val htmlDescription = rule.htmlDescription()?.lines();
```
leveretka, line 42: I think it is safe to use `!!` here and get rid of nullability. As we don't expect decsriotion to be null and if it is it wili fail accordingly

leveretka (approved):
Looks good, have a minor suggestion to improve nullability handling

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinRulesDefinitionTest.kt
antonioaversa: Indeed, I have also removed the unnecessary `;`.

antonioaversa:
@leveretka I have addressed the comment.
Can you please merge? I cannot do myself, since `master` is locked.
Once this is merged, I will rebase the https://github.com/SonarSource/sonar-kotlin/pull/523
Thanks

PR #541, opened 2025-01-30
SONARKT-495 Remove compatibility with SonarQube versions below 10.8

leveretka:
[SONARKT-495](https://sonarsource.atlassian.net/browse/SONARKT-495)

[SONARKT-495]: https://sonarsource.atlassian.net/browse/SONARKT-495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -167,7 +167,7 @@ qa_plug_pub_lin_task:
     GRADLE_TASK: ":its:plugin:test"
     matrix:
       - SQ_VERSION: "DEV"
-      - SQ_VERSION: "LATEST_RELEASE[9.9]"
+      - SQ_VERSION: "LATEST_RELEASE[25.1]"
       - SQ_VERSION: "LATEST_RELEASE"
   <<: *LINUX_6_CPU_12G_JAVA_17
   <<: *GRADLE_ITS_TEMPLATE
@@ -175,7 +175,7 @@ qa_plug_pub_lin_task:
 qa_plug_pub_win_task:
   <<: *QA_TASK_FILTER
   env:
-    SQ_VERSION: LATEST_RELEASE[9.9]
+    SQ_VERSION: LATEST_RELEASE
     GRADLE_TASK: ":its:plugin:test"
     ITS_PROJECT: "plugin"
   <<: *WINDOWS_16_CPU_32G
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -167,15 +167,15 @@ qa_plug_pub_lin_task:
...
+      - SQ_VERSION: "LATEST_RELEASE[25.1]"
       - SQ_VERSION: "LATEST_RELEASE"
   <<: *LINUX_6_CPU_12G_JAVA_17
   <<: *GRADLE_ITS_TEMPLATE
 
 qa_plug_pub_win_task:
   <<: *QA_TASK_FILTER
   env:
-    SQ_VERSION: LATEST_RELEASE[9.9]
+    SQ_VERSION: LATEST_RELEASE
```
leveretka, line 178: This doesn't make sense to test again the LTA

antonioaversa: I am not sure how these tags work exactly. Don't `LATEST_RELEASE[25.1]` and `LATEST_RELEASE` currently point to the same thing? 
Shouldn't we test against `LATEST_RELEASE[10]` or `LATEST_RELEASE[24.12]` instead?

leveretka: As far as I understand LTA is 2025.1, and we should test with it (while the analyzer won't be updated, we could potentially do backport there)

The latest releases is probably the most important test to validate our functionality(its still fresh and it's more stable than DEV) At the moment it should also point to the LTA, but once the next version is released it will point there.

The DEV is actually the most relevant one as analyzer will be shipped in the next versions of SQ, not already released ones. However, DEV could potentially be unstable and seeing that something works on the Latest release and is broken in DEV is very helpful.

Indeed, you're right that for some time LTA and LATEST RELEASE are the same, so to simplify things we can potentially drop LTA and olny test against the latest release and DEV. Wdyt?

antonioaversa: I guess the LTA was there to test the behavior in case of hotfixes that need to be backported to LTA. But that the scenario seems to have low probability and I agree with you that it is OK to drop it, and make the CI a bit shorter/less computationally intensive.

antonioaversa (approved):
Thanks for the answer to the clarification question. LGTM

PR #560, opened 2025-02-05
SONARKT-542 Fix CPD with string templates

antonioaversa:
[SONARKT-542](https://sonarsource.atlassian.net/browse/SONARKT-542)

See [here](https://community.sonarsource.com/t/line-duplication-detection-in-kotlin-file-way-too-aggressive/82880/23?u=antonio.aversa) for a complete analysis of the issue.

The adopted solution is to treat the entire string template as a single `LITERAL` token. In the previous solutions, each `KtStringTemplateEntry` emitted a `LITERAL` token.

### Downsides
This may not be perfect, as we may not detect **duplication patterns in expressions within interpolations of a string template**.
For example, the following string:
```kotlin
var x5 = """ 
    a ${<interpolation1>} b 
    c ${<interpolation2>} d 
    ${<interpolation3>} f""".trimIndent()
```
We would have a single `LITERAL` for the entire string template: `varx5=LITERAL`.
Therefore, we will not see duplications between `<interpolation1>`, `<interpolation2>`, and `<interpolation3>`.

We will also not see **equivalent interpolation structures**.
For example, given the following two strings:
```kotlin
var x1 = """"
    a $b c $d e $f ... $something
    """"
var x2 = """"
    a $b c $d e $f ... $somethingElse
    """"
```
We will see just two `LITERAL`: one for `x1` and one for `x2`, instead of a sequence of `LITERAL` which would be the same, up to `$something`/`$somethingElse`

For both scenarios mentioned above, the probability that a duplication would be detected (i.e. at least 100 equal tokens on at least 10 different lines) seems close to zero, so it's better to treat the entire template as a literal to avoid issues like the one explained in [the community post](https://community.sonarsource.com/t/line-duplication-detection-in-kotlin-file-way-too-aggressive/82880/23?u=antonio.aversa). 

[SONARKT-542]: https://sonarsource.atlassian.net/browse/SONARKT-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/cpd/CopyPasteDetectorTest.kt
```diff
@@ -17,11 +17,14 @@
 package org.sonarsource.kotlin.plugin.cpd
 
 import com.intellij.openapi.util.Disposer
+import net.bytebuddy.utility.JavaConstant.Dynamic
```
leveretka, line 20: It seems to me, you don't need this import. It's not raised because it's test.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/cpd/CopyPasteDetectorTest.kt
```diff
@@ -33,6 +36,7 @@ import org.sonarsource.kotlin.plugin.DummyReadCache
 import org.sonarsource.kotlin.plugin.DummyWriteCache
 import org.sonarsource.kotlin.api.checks.InputFileContextImpl
 import org.sonarsource.kotlin.api.frontend.Environment
+import org.sonarsource.kotlin.testapi.DummyInputFile
```
leveretka, line 39: This import is also odd. It's not raised because it's test.

leveretka (approved):
In general, the approach looks good, thanks for having a look and fixing this.

I've just noticed 2 unused imports in the test.

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/cpd/CopyPasteDetectorTest.kt
antonioaversa: Indeed, good catch! Fixed

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/cpd/CopyPasteDetectorTest.kt
antonioaversa: Correct, fixed as well

antonioaversa:
@leveretka `master` is locked, so I'd let you merge the PR

PR #566, opened 2025-02-11
SONARKT-552 Migrate DelegationPatternCheck to kotlin-analysis-api

Godin:
[SONARKT-552](https://sonarsource.atlassian.net/browse/SONARKT-552)

Part of 
SONARKT-532

[SONARKT-552]: https://sonarsource.atlassian.net/browse/SONARKT-552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/ruling/src/test/resources/expected/kotlin/kotlin/kotlin-S6514.json
+++ b/its/ruling/src/test/resources/expected/kotlin/kotlin/kotlin-S6514.json
@@ -85,6 +85,9 @@
 "kotlin-kotlin-project:sources/kotlin/kotlin/idea/idea-analysis/src/org/jetbrains/kotlin/idea/caches/project/IdeaModuleInfos.kt": [
 443
 ],
+"kotlin-kotlin-project:sources/kotlin/kotlin/idea/idea-jvm/src/org/jetbrains/kotlin/idea/scratch/output/InlayScratchOutputHandler.kt": [
+45
+],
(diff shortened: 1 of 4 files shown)


its/ruling/src/test/resources/expected/kotlin/kotlin/kotlin-S6514.json
```diff
@@ -1,6 +1,6 @@
 {
```
leveretka, line 1: Issues disappeared from the Kotlin compiler are related to the fact, that we only pass Kotlin files to the analysis and don't provide classpath.

The issues disappear because for some parameters we can't resolve types (receive KaErrorType) and that's why we can't guarantee that there was a match of method signatures. The types that are unresolved are represented by classes written in Java (for example KtExpression).

So far I think we can accept this change as this is related to the analysis misconfiguration and not the rule behavior.

leveretka: Previosly an empty String was returned as a fully qualified name of an Error type. Since `""` is equal to `""`, the issue appeared by accident and by accident those issues happened to be TP.

Before we introduced generation of Binding context in all cases, these issues weren't triggered as types weren't resolved at all.

leveretka (approved):

PR #572, opened 2025-02-20
SONARKT-569 Add support for WebViews to S4830

antonioaversa:
[SONARKT-569](https://sonarsource.atlassian.net/browse/SONARKT-569)

[SONARKT-569]: https://sonarsource.atlassian.net/browse/SONARKT-569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
+++ b/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
@@ -30,7 +30,6 @@ import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
 import org.jetbrains.kotlin.psi.KtThrowExpression
 import org.jetbrains.kotlin.psi.KtTreeVisitorVoid
 import org.jetbrains.kotlin.psi.KtTryExpression
-import org.jetbrains.kotlin.psi.KtVisitorVoid
 import org.jetbrains.kotlin.psi.KtWhenExpression
 import org.sonar.check.Rule
 import org.sonarsource.kotlin.api.checks.AbstractCheck
--- a/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
+++ b/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
@@ -143,8 +143,10 @@ class ServerCertificateCheck : AbstractCheck() {
         var potentialBranching: Boolean = false
 
         override fun visitCallExpression(expression: KtCallExpression) {
-            proceedFound = proceedFound || androidSslErrorHandlerProceedFunMatcher.matches(expression)
-            cancelFound = cancelFound || androidSslErrorHandlerCancelFunMatcher.matches(expression)
+            when {
+                androidSslErrorHandlerProceedFunMatcher.matches(expression) -> proceedFound = true
+                androidSslErrorHandlerCancelFunMatcher.matches(expression) -> cancelFound = true
+            }
             super.visitCallExpression(expression)
         }
 
--- a/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
+++ b/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
@@ -16,7 +16,6 @@
  */
 package org.sonarsource.kotlin.checks
 
-import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.psi.KtBinaryExpression
@@ -29,6 +28,7 @@ import org.jetbrains.kotlin.psi.KtLoopExpression
 import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
 import org.jetbrains.kotlin.psi.KtThrowExpression
+import org.jetbrains.kotlin.psi.KtTreeVisitorVoid
 import org.jetbrains.kotlin.psi.KtTryExpression
 import org.jetbrains.kotlin.psi.KtVisitorVoid
 import org.jetbrains.kotlin.psi.KtWhenExpression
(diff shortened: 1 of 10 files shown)


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -104,6 +125,43 @@ class ServerCertificateCheck : AbstractCheck() {
...
+
+        var foundProceedCall: Boolean = false
+        var foundCancelCall: Boolean = false
+        var potentialBranching: Boolean = false
+
+        override fun visitCallExpression(expression: KtCallExpression) {
+            val isProceedCall = androidSslErrorHandlerProceedFunMatcher.matches(expression)
+            val isCancelCall = androidSslErrorHandlerCancelFunMatcher.matches(expression)
+            foundProceedCall = foundProceedCall || isProceedCall
+            foundCancelCall = foundCancelCall || isCancelCall
```
leveretka, line 147: here we always try to match, both proceed and cancel and always override the foundProceedCall and foundCancelCall.

It seems to me we can inline 
```suggestion
            foundProceedCall = foundProceedCall || androidSslErrorHandlerProceedFunMatcher.matches(expression)
            foundCancelCall = foundCancelCall || androidSslErrorHandlerCancelFunMatcher.matches(expression)
```

leveretka: Also since these calls can't be matched together, we can refactor to only match one. I'll let you decide how is better to refactor it.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -104,6 +125,43 @@ class ServerCertificateCheck : AbstractCheck() {
         }
     }
 
+    private class AndroidWebViewCandidateVisitor : KtVisitorVoid() {
```
leveretka, line 128: That's FYI, Looks like for this rule we implemented recursive visit function by ourselves, however, there's a `KtTreeVisitorVoid`, which is recursive. No need to refactor this check, just FYI for the future rules.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -104,6 +125,43 @@ class ServerCertificateCheck : AbstractCheck() {
...
+
+        override fun visitCallExpression(expression: KtCallExpression) {
+            val isProceedCall = androidSslErrorHandlerProceedFunMatcher.matches(expression)
+            val isCancelCall = androidSslErrorHandlerCancelFunMatcher.matches(expression)
+            foundProceedCall = foundProceedCall || isProceedCall
+            foundCancelCall = foundCancelCall || isCancelCall
+        }
+
+        override fun visitExpression(expression: KtExpression) {
+            potentialBranching = potentialBranching ||
```
egon-okerman-sonarsource, line 151: Educational question: does this cover all branching?

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -17,53 +17,66 @@
...
 
     override fun visitNamedFunction(function: KtNamedFunction, kotlinFileContext: KotlinFileContext) {
-        if (function.belongsToTrustManagerClass()
+        val javaCryptographyExtensionCandidate = javaCryptographyExtensionFunMatchers.any { it.matches(function) }
             && !function.callsCheckTrusted()
             && !function.throwsCertificateExceptionWithoutCatching()
-        ) {
+        val androidWebViewCandidate = androidWebViewFunMatchers.any { it.matches(function) }
+            && function.callsProceedUnconditionally()
+        if (javaCryptographyExtensionCandidate || androidWebViewCandidate) {
```
leveretka, line 68: Here we match function to both FunMatchers regardless of the result. I'd suggest extracting refactoring the code so we don't match function a second time. As an idea it can look like this:

```
when {
            javaCryptographyExtensionFunMatchers.any { it.matches(function) } ->
                if (!function.callsCheckTrusted() && !function.throwsCertificateExceptionWithoutCatching())
                    kotlinFileContext.reportIssue(function.nameIdentifier ?: function,
                        "Enable server certificate validation on this SSL/TLS connection.")
            androidWebViewFunMatchers.any { it.matches(function) } ->
                if (function.callsProceedUnconditionally()) {
                    kotlinFileContext.reportIssue(function.nameIdentifier ?: function,
                        "Enable server certificate validation on this SSL/TLS connection.")
                }
        }
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -72,15 +85,23 @@ class ServerCertificateCheck : AbstractCheck() {
...
 
+    /*
+     * The following heuristic is used to determine if the call to "proceed" is unconditional:
+     * proceed is called, cancel is never called, and there are no branching statements in the
+     * function (no matter if around the proceed or not).
+     */
+    private fun KtNamedFunction.callsProceedUnconditionally(): Boolean {
+        val visitor = AndroidWebViewCandidateVisitor()
+        this.acceptRecursively(visitor)
+        return visitor.foundProceedCall && !visitor.foundCancelCall && !visitor.potentialBranching
```
leveretka, line 102: That's more a matter of taste, what do you think about this way:

`with(visitor) { foundProceedCall && !foundCancelCall && !potentialBranching }`

leveretka (changes requested):
Nice work! I see we can have a few improvements here.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
antonioaversa: Yeah, good question - the answer is: all the ones I could think of :-)
I find unlikely that somebody will find a way to have branching, have `proceed` and not have any cancel, in other ways... but I would not underestimate people creativity :-)
Anyway, unless I am missing some common scenarios, the risk for FP seems close to zero to me. 
If we are concerned about this heuristic we can always remove it go for the Option 3 we discussed (only simple `proceed` in body), but that seems too much on the side of safety, at the expense of reach.
WDYT?

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
antonioaversa: I have changed the code to use`KtTreeVisitorVoid` everywhere, so that I could remove the logic to traverse descendants. Except for the call to `super`, it was a drop-in replacement.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
antonioaversa: Yes, nice! I used it on the other visitor as well, for consistency.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
antonioaversa: Indeed, that's unnecessary computation.
I came up with something slightly different to avoid repeating reporting logic twice:
```
    override fun visitNamedFunction(function: KtNamedFunction, kotlinFileContext: KotlinFileContext) {
        val match = when {
            javaCryptographyExtensionFunMatchers.any { it.matches(function) } ->
                !function.callsCheckTrusted() && !function.throwsCertificateExceptionWithoutCatching()
            androidWebViewFunMatchers.any { it.matches(function) } ->
                function.callsProceedUnconditionally()
            else -> false
        }
        if (match) {
            kotlinFileContext.reportIssue(function.nameIdentifier ?: function,
                "Enable server certificate validation on this SSL/TLS connection.")
        }
    }
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -72,15 +85,23 @@ class ServerCertificateCheck : AbstractCheck() {
         return visitor.callsCheckTrusted()
     }
 
-    /*
-     * Returns true only when the function throws a CertificateException without a catch against it.
-     */
```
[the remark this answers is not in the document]
antonioaversa: Sure, I removed it because it seemed redundant - it basically says the same thing as the name of the function, without any further detail.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/ServerCertificateCheck.kt
```diff
@@ -16,59 +16,84 @@
...
 
 @Rule(key = "S4830")
 class ServerCertificateCheck : AbstractCheck() {
 
     override fun visitNamedFunction(function: KtNamedFunction, kotlinFileContext: KotlinFileContext) {
-        if (function.belongsToTrustManagerClass()
-            && !function.callsCheckTrusted()
-            && !function.throwsCertificateExceptionWithoutCatching()
-        ) {
+        val match = when {
```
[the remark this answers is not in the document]
leveretka: Another option could be to use `ifTrue {}`

leveretka (approved):
Looks good! Let's just wait till the K2 is fully merged before going to master

PR #583, opened 2025-03-06
SONARKT-600 Rule S6474: Using remote artifacts without authenticity and integrity checks is security-sensitive

antonioaversa:
[SONARKT-600](https://sonarsource.atlassian.net/browse/SONARKT-600)
RSPEC update: https://github.com/SonarSource/rspec/pull/4723

[SONARKT-600]: https://sonarsource.atlassian.net/browse/SONARKT-600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/KotlinGradleSensor.kt
```diff
@@ -97,4 +99,25 @@ class KotlinGradleSensor(
             }
         }
     }
+
+    private fun checkForMissingVerificationMetadata(rootDirFile: File, sensorContext: SensorContext) {
```
leveretka, line 103: It seems to me this method has much duplication with the one checkoing for missing settings. Do you think we can we can extract some common logic here, or it will make things less clear. I don't have opinion here, wdyt?

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/KotlinGradleSensor.kt
```diff
@@ -97,4 +99,25 @@ class KotlinGradleSensor(
...
+            val project = sensorContext.project()
+
+            with(sensorContext) {
+                newIssue()
+                    .forRule(missingSettingsRuleKey)
+                    .at(
+                        newIssue()
+                            .newLocation()
+                            .on(project)
+                            .message("""Create a "verification-metadata.xml" file to verify these dependencies against a known checksum or signature.""")
```
leveretka, line 117: I remember that for hotspots we usually had messages in format "Make sure ... is safe here..." or something similar. I am again not sure here, so I'll let you clarify this with @pierre-loup-tristant-sonarsource

sonar-kotlin-gradle/src/test/java/org/sonarsource/kotlin/gradle/KotlinGradleSensorTest.kt
```diff
@@ -147,6 +148,108 @@ internal class KotlinGraldeSensorTest : AbstractSensorTest() {
...
+    @Test
+    fun test_missing_verification_metadata_rule_is_trigger_when_Kotlin_settings_is_used_without_verification_metadata() {
+        mockkStatic("org.sonarsource.kotlin.gradle.KotlinGradleCheckListKt")
+        every { KOTLIN_GRADLE_CHECKS } returns listOf(MissingVerificationMetadataCheck::class.java)
+
+        val settings = MapSettings()
+        settings.setProperty(GRADLE_PROJECT_ROOT_PROPERTY, baseDir.toRealPath().toString())
+        context.setSettings(settings)
+
+        addSettingsKtsFile()
```
leveretka, line 160: Since the presence of `settings` file is not the part of the rule logic, maybe we can simplify the test and omit adding `build.gradle` or `settings.gradle`.

On the other hand, I'd probably check when the property `GRADLE_PROJECT_ROOT_PROPERTY` is not set, to make sure we don't report in that case

leveretka (commented):
The implementation looks good.

Had a few tests and style comments and let's double-check the message

sonar-kotlin-gradle/src/test/java/org/sonarsource/kotlin/gradle/KotlinGradleSensorTest.kt
antonioaversa: Yes: I have removed this test and the following (which were testing for Kotlin DSL and Groovy settings), and added a test on `GRADLE_PROJECT_ROOT_PROPERTY`. While the `GRADLE_PROJECT_ROOT_PROPERTY` check was already covered by the UTs for `checkForMissingGradleSettings`, an additional test ensures that if `checkForMissingVerificationMetadata` is removed from the following `if`
```kotlin
        sensorContext.config()[GRADLE_PROJECT_ROOT_PROPERTY].ifPresent {
            checkForMissingGradleSettings(File(it), sensorContext)
            checkForMissingVerificationMetadata(File(it), sensorContext)
        }
```
 while `checkForMissingGradleSettings` is not, there is a test turning red. 

So adding the test is relevant.

leveretka (approved):
Looks good!

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/KotlinGradleSensor.kt
pierre-loup-tristant-sonarsource: You are right @margarita-nedzelska-sonarsource. Here is a new suggestion for this case:
"Dependencies are not verified because the "verification-metadata.xml" file is missing. Make sure it is safe here."

sonar-kotlin-gradle/src/main/java/org/sonarsource/kotlin/gradle/checks/MissingVerificationMetadataCheck.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.sonarsource.kotlin.gradle.checks
+
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.sonar.check.Rule
+import org.sonarsource.kotlin.api.checks.AbstractCheck
+import org.sonarsource.kotlin.api.frontend.KotlinFileContext
+
+private const val disableVerificationMetadata = "disableDependencyVerification"
+private const val message = """Remove this call to "disableDependencyVerification()" so that dependencies are verified."""
```
pierre-loup-tristant-sonarsource, line 25:
```suggestion
private const val message = """This call disables dependencies verification. Make sure it is safe here."""
```

pierre-loup-tristant-sonarsource:
Updated issue messages is defined in https://github.com/SonarSource/rspec/pull/4758/files)

PR #598, opened 2025-03-21
SONARKT-618 Rule S7410: Keyboard cache should be disabled for password inputs

antonioaversa:
[SONARKT-618](https://sonarsource.atlassian.net/browse/SONARKT-618)

[SONARKT-618]: https://sonarsource.atlassian.net/browse/SONARKT-618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kotlin-checks-test-sources/src/main/kotlin/androidx/compose/material/OutlinedTextField.kt
```diff
@@ -0,0 +1,30 @@
+package androidx.compose.material
+
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.foundation.text.input.TextFieldState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.input.VisualTransformation
+
+// Partial stubbing of parameters
```
antonioaversa, line 8: I added these comments about partial stubbing so that it will be easier to recognize for which methods the stub has missing parameters, and add if needed.
I couldn't have complete stubs with all parameters because many functions and constructors take a large number of parameters, that would have forced me to mock way more classes.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/AndroidKeyboardCacheOnPasswordInputCheck.kt
```diff
@@ -0,0 +1,169 @@
...
+            kotlinFileContext.reportIssue(
+                keyboardOptionsArgument,
+                WITH_KEYBOARD_OPTIONS_MESSAGE,
+                secondaryLocations = secondaryLocations,
+            )
+        }
+    }
+
+    private fun KaFunctionCall<*>.getPasswordVisualTransformationConstructorOrNull() : KtExpression? =
+        argumentMapping.entries.singleOrNull { it.value.symbol.name == visualTransformationParamName }
```
antonioaversa, line 110: Is this the good way of matching a parameter (i.e. by name)?
I was wondering if there are scenarios where the original parameter name is not available (e.g. has somehow been transformed into something like `argN` by minification or obfuscation), since the type is an Android type.
If that can happen, I can match parameter by type, assuming that there is a single `VisualTransformation` parameter in the signature... which is currently the case, but may not be the case in the future.

leveretka: this happens if the original is written in Java. Java parameter names are not preserved. So far we only relied on the parameter order. It is preserved in the symbol

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/AndroidKeyboardCacheOnPasswordInputCheck.kt
```diff
@@ -0,0 +1,169 @@
...
+        // e.g. KeyboardOptions(keyboardType = KeyboardType.Ascii)
+        // no keyboardType argument means KeyboardType.Unspecified, that is cache enabled
+        return@withKaSession keyboardOptionsConstructorFunMatcher.matches(expression) &&
+            (isCacheEnabledKeyboardTypeOrNull(expression) ?: true)
+    }
+
+    private fun isKeyboardOptionsWithCacheEnabledReferenceExpression(expression: KtReferenceExpression): Boolean = withKaSession {
+        val variableAccessCall = expression.resolveToCall()?.successfulVariableAccessCall() ?: return@withKaSession false
+        val symbol = (variableAccessCall as? KaSimpleVariableAccessCall)?.symbol ?: return@withKaSession false
+        val value = (symbol.psi as? KtProperty)?.takeIf { !it.isVar }?.initializer
```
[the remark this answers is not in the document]
leveretka: Fine for me. Maybe we can reconsider this later for the security rules, as we want a higher reach in such rules. For now let's keep it like this

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/AndroidKeyboardCacheOnPasswordInputCheck.kt
```diff
@@ -0,0 +1,169 @@
...
+            ?: (symbol.psi as? KtParameter)?.defaultValue
+            ?: return@withKaSession false
+        return@withKaSession isKeyboardOptionsWithCacheEnabled(value)
+    }
+
+    private fun isCacheEnabledKeyboardTypeOrNull(expression: KtCallExpression) : Boolean? = withKaSession {
+        val resolvedCall = expression.resolveToCall()?.successfulFunctionCallOrNull() // Can be constructor or normal function call
+        return@withKaSession resolvedCall?.argumentMapping?.entries?.singleOrNull {
```
leveretka, line 159: Here and in other places `return` is also unnecessary as the last expression is returned

leveretka (approved):
Nice and clean implementation. Just had a few cosmetic comments

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/AndroidKeyboardCacheOnPasswordInputCheck.kt
```diff
@@ -0,0 +1,169 @@
...
+            it.value.symbol.name == keyboardTypeParamName
+        }?.key?.let(::isCacheEnabledKeyboardType)
+    }
+
+    private fun isCacheEnabledKeyboardType(expression: KtExpression): Boolean = withKaSession {
+        return@withKaSession expression is KtDotQualifiedExpression &&
```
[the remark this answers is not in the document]
antonioaversa: Oh, I forgot about that. I removed the explicit `return` and made the overall expression an inliner.
Looks better!

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/AndroidKeyboardCacheOnPasswordInputCheck.kt
antonioaversa: Yes, changed and inlined `resolvedCall`, resulting in a single flow of `?.` operators.

PR #601, opened 2025-03-26
SONARKT-639 Rule S5344: Passwords should not be stored in plaintext or with a fast hashing algorithm

antonioaversa:
[SONARKT-639](https://sonarsource.atlassian.net/browse/SONARKT-639)

RSPEC: https://github.com/SonarSource/rspec/pull/4803/files
RSPEC Page: https://sonarsource.github.io/rspec/#/rspec/S5344

[SONARKT-639]: https://sonarsource.atlassian.net/browse/SONARKT-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PasswordPlaintextFastHashingCheck.kt
```diff
@@ -0,0 +1,128 @@
...
+            kotlinFileContext.reportIssue(iterationCountExpression, MESSAGE.format(minIteration), secondaryLocations)
+        }
+    }
+
+    private fun KtExpression.predictRuntimeValueFromParameterDefault(): KtExpression = withKaSession {
+        resolveToCall()
+            ?.successfulVariableAccessCall()
+            ?.symbol
+            ?.psi
+            ?.let { it as? KtParameter }
```
antonioaversa, line 124: Is this the best way to go from a Symbol (which in this case is a `KaValueParameterSymbol`) to its `KtParameter`, in order to get its default value?
The symbol itself, when cast to `KaValueParameterSymbol`, has a `hasDefaultValue` property, but apparently no direct way to jump to the `KtExpression` of such default value.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PasswordPlaintextFastHashingCheck.kt
```diff
@@ -0,0 +1,127 @@
...
+            val secondaryLocations = mutableListOf(SecondaryLocation(kotlinFileContext.textRange(algorithmValueExpression)))
+            if (iterationCountValueExpression.textRange != iterationCountExpression.textRange) {
+                secondaryLocations.add(SecondaryLocation(kotlinFileContext.textRange(iterationCountValueExpression)))
+            }
+
+            kotlinFileContext.reportIssue(iterationCountExpression, MESSAGE.format(minIteration), secondaryLocations)
+        }
+    }
+
+    private fun KtExpression.predictRuntimeValueFromParameterDefault(): KtExpression = withKaSession {
```
antonioaversa, line 118: I was thinking of extracting this method into general utilities, but its use is very specific and very contextual, and in general could make false prediction since non-val parameters can be reassigned before their use, within the method.

So in the end I left it as a `private fun`.

leveretka (approved):
Implementation looks good

See a few comments to improve

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PasswordPlaintextFastHashingCheck.kt
```diff
@@ -0,0 +1,127 @@
...
+class PasswordPlaintextFastHashingCheck : CallAbstractCheck() {
+
+    override val functionsToVisit = listOf(generateSecretFunMatcher)
+
+    override fun visitFunctionCall(
+        callExpression: KtCallExpression,
+        resolvedCall: KaFunctionCall<*>,
+        kotlinFileContext: KotlinFileContext,
+    ) = withKaSession {
+        generateSecretFunMatcher.matches(resolvedCall) || return@withKaSession
```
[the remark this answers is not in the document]
antonioaversa: Indeed, that was also visible from coverage. I removed the line.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PasswordPlaintextFastHashingCheck.kt
antonioaversa: Interesting, I will keep that in mind! 
In the specific case at hand, I cannot see it happening, except maybe if the iteration count or the algorithm is defined in a property a base class, which resides in a different package for which we don't have sources. However, in that case, no syntactic approach would work either: if we don't have the AST, we cannot get the default value via it.
Moreover such scenario has virtually 0 probability of happening, and in the worst case we would have a False Negative. So I would conclude that the approach is safe.

PR #602, opened 2025-03-28
SONARKT-597 Fix NPE when If statement doesn't contain "then" branch

leveretka:


Godin (changes requested):
According to

```
grep "\.then\!\!" -R sonar-kotlin-checks
```

there are other occurrences of similar non-null assertion.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/LiftReturnStatementCheck.kt
```diff
@@ -32,7 +32,7 @@ class LiftReturnStatementCheck : AbstractCheck() {
 
     override fun visitIfExpression(expression: KtIfExpression, context: KotlinFileContext) {
         // `then` branch of an `if` statement can never be null
-        val thenBranch = expression.then!!
+        val thenBranch = expression.then ?: return
```
Godin, line 35: Comment does not match code anymore 😉

Godin (changes requested):

Godin (approved):

Godin (approved):

PR #624, opened 2025-06-11
SONARKT-670 Use "sonar.scanner.skipJreProvisioning" in integration tests

Godin:


leveretka:
We can accept the cahnges. Because. Kotli compiler relies on the `com.sun.jdi` package. This package seems to be present in the JRE taken from our CI, but not present in the JRE, provided by SQ. So technically this occurs due to the fact that we don't add dependencies explicitly in the ruling test for the Kotlin compiler. (The workaround for users will probably be to provide `jdkHome` property)

I will also double-check if that is expected and what else could be missing. However, I don't think this issue stops us from merging as it has nothing to do with Kotlin analyzer.

PR #637, opened 2025-09-03
SC-27381 Add shadow scan github workflow

johann-beleites-sonarsource:


jonas-wielage-sonarsource:
Will be resumed in https://sonarsource.atlassian.net/browse/SONARKT-742

PR #657, opened 2026-01-28
SONARKT-259 Use stricter method resoultion to reduce incorrect resolutions…

johann-beleites-sonarsource:
… when semantics are missing.


jonas-wielage-sonarsource:
Assigned ticket is put back into TODO, another PR will be reopened in case the effort is resumed.

PR #688, opened 2026-03-25
SONARKT-25 Update ruling projects to use Gradle scanner

petertrr:
This PR bumps some ruling projects to recent versions and specifies language version explicitly in the ruling tests.

* `kotlin_compiler` is not updated, because it's a large project that already takes about half of the time on CI. It's better suited for Peachee than ruling
* `okio` is not updated, because this is a multiplatform project with a complex structure, which is more suited for Peachee than ruling


petertrr:
Will be rebased and cleaned up later

PR #695, opened 2026-04-07
SONARKT-736 SONARKT-716 Migrate rule metadata update to cloud-native rule-api plugin; update rule metadata

petertrr:
Replace custom FetchRuleMetadata/GitHubApi buildSrc tasks with the org.sonarsource.cloud-native.rule-api convention plugin from build-logic/common. Run rule metadata update and enable it by default in AutomateRelease.yml.

Part of SONARKT-713


build.gradle.kts
```diff
@@ -293,14 +290,11 @@ artifactory {
...
+    finalizedBy(":sonar-kotlin-plugin:ruleApiGenerateRuleKotlin")
 }
 
 tasks.register<CreateKotlinGradleRuleStubsTask>("setupGradleRuleStubs") {
     group = "Rules"
     description = "Generate required stubs for a new Kotlin Gradle DSL rule"
-    finalizedBy(tasks.getByPath(":generateRuleMetadata"))
+    finalizedBy(":sonar-kotlin-plugin:ruleApiGenerateRuleKotlin")
 }
-
```
petertrr, line 308: rule-metadata tasks from gradle-modules are not completely equivalent to these ones: the old ones were able to fetch a list of GH PRs labelled with a language and allow to interactively choose the branch to fetch from. Personally, I've never had a problem with specifying branch manually, but OTOH this functionality can be eventually moved into gradle-modules.

jonas-wielage-sonarsource: Sure we can move this functionality if it's not too much effort. However I also never had problems specifying the branch and Koushik was also able to easily do it without knowing the plugin before.

sonar-kotlin-plugin/sonarpedia.json
```diff
@@ -3,7 +3,7 @@
   "languages": [
     "KOTLIN"
   ],
-  "latest-update": "2025-07-12T19:40:22.641523Z",
+  "latest-update": "2026-03-31T11:50:55.099110304Z",
```
jonas-wielage-sonarsource: 😮

jonas-wielage-sonarsource (approved):
🫡

sonar-kotlin-plugin/sonarpedia.json
petertrr: There were some partial updates to metadata even after this date, so it's not _that_ bad :D

PR #696, opened 2026-04-07
SONARKT-719 S6518 should not raise when get or set is called with named parameters

matthew-reid-sonarsource:
Part of


petertrr (approved):
LGTM! One nitpick: convention is to name a PR with exact ticket title

PR #701, opened 2026-04-13
SONARKT-739 S6518 Add class exceptions for common user errors where operator is provided via java interop

matthew-reid-sonarsource:
Part of

### DIFF
--- a/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/IndexedAccessCheck.kt
+++ b/sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/IndexedAccessCheck.kt
@@ -18,10 +18,10 @@ package org.sonarsource.kotlin.checks
 
 import org.jetbrains.kotlin.analysis.api.resolution.KaFunctionCall
 import org.jetbrains.kotlin.analysis.api.resolution.symbol
+import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolOrigin
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
-import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtSuperExpression
 import org.sonar.check.Rule
 import org.sonarsource.kotlin.api.checks.CallAbstractCheck
@@ -99,11 +99,8 @@ class IndexedAccessCheck : CallAbstractCheck() {
      * with the `operator` keyword in Kotlin source).
      */
     private fun isJavaInteropOperator(resolvedCall: KaFunctionCall<*>): Boolean {
-        val receiverType = getReceiverTypeFqn(resolvedCall)
-        // Kotlin-declared types always have Kotlin-defined operators, not Java interop
-        if (receiverType != null && receiverType.startsWith("kotlin.")) return false
-        // If the function is declared in Kotlin source, it's not a Java interop operator
-        return resolvedCall.symbol.psi !is KtNamedFunction
+        return resolvedCall.symbol.origin == KaSymbolOrigin.JAVA_SOURCE
+            || resolvedCall.symbol.origin == KaSymbolOrigin.JAVA_LIBRARY
     }
 
     /**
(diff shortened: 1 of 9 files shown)


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/IndexedAccessCheck.kt
```diff
@@ -45,6 +85,35 @@ class IndexedAccessCheck : CallAbstractCheck() {
...
+     * Checks whether the resolved function is a Java interop operator (i.e., a Java method
+     * that Kotlin treats as an operator via interop, rather than a function explicitly declared
+     * with the `operator` keyword in Kotlin source).
+     */
+    private fun isJavaInteropOperator(resolvedCall: KaFunctionCall<*>): Boolean {
+        val receiverType = getReceiverTypeFqn(resolvedCall)
+        // Kotlin-declared types always have Kotlin-defined operators, not Java interop
+        if (receiverType != null && receiverType.startsWith("kotlin.")) return false
+        // If the function is declared in Kotlin source, it's not a Java interop operator
+        return resolvedCall.symbol.psi !is KtNamedFunction
```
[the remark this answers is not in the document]
petertrr: @matthew-reid-sonarsource can you add a test with a method from external dependency to validate this comment? I believe the `KaJavaMethodSymbol` doesn't exist (at least I wasn't able to find it), but the concern itself seems reasonable

petertrr: Also, did you look at `resolvedCall.symbol.origin`? If it's set by the compiler reliably enough, it can be more suitable than `symbol.psi`.

petertrr (changes requested):
Good progress, but more examples with non-stdlib dependencies are needed.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/IndexedAccessCheck.kt
matthew-reid-sonarsource: 👍 Looks like it is set reliably and is definitely a better choice here, I went ahead and updated.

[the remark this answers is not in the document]
matthew-reid-sonarsource: Confirmed via tests that this was a real issue. Just pushed up a fix

sonar-kotlin-plugin/build.gradle.kts
```diff
@@ -219,7 +219,7 @@ val distTask = tasks.register<ProGuardTask>("dist") {
     outjars("build/libs/sonar-kotlin-plugin.jar")
     configuration("proguard.txt")
     doLast {
-        enforceJarSizeAndCheckContent(file("build/libs/sonar-kotlin-plugin.jar"), 53_300_000L, 53_800_000L)
+        enforceJarSizeAndCheckContent(file("build/libs/sonar-kotlin-plugin.jar"), 53_300_000L, 54_000_000L)
```
petertrr: Nitpick: we keep the allowed range at 500k, so please bump the lower boundary too

petertrr (approved):
LGTM!

sonar-kotlin-plugin/build.gradle.kts
matthew-reid-sonarsource: Ah, gotcha. I'll go update 👍

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/IndexedAccessCheck.kt
```diff
@@ -45,6 +85,32 @@ class IndexedAccessCheck : CallAbstractCheck() {
...
+    /**
+     * Checks whether the receiver type of the resolved function is in the allow-list
+     * of Java types where indexed access operator syntax is idiomatic.
+     */
+    private fun isAllowedJavaInteropType(resolvedCall: KaFunctionCall<*>): Boolean = withKaSession {
+        if (getReceiverTypeFqn(resolvedCall) in JAVA_INTEROP_ALLOWED_TYPES) return true
+        // Any List or Map implementation supports idiomatic indexed access
+        val type = (resolvedCall.dispatchReceiver ?: resolvedCall.extensionReceiver)?.type ?: return false
+        return type.isSubtypeOf(LIST_CLASS_ID) || type.isSubtypeOf(MAP_CLASS_ID)
+    }
```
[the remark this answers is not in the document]
matthew-reid-sonarsource: Fixed

PR #709, opened 2026-04-17
SONARKT-752 Exclude bouncycastle from the build classpath

petertrr:
Part of


petertrr:
Closing in favor of #710

PR #731, opened 2026-06-08
SONARKT-769 Add telemetry for Kotlin parse failures

M1ke-Gru:
This is the ticket description copy from Jira.
# WHY
We want visibility into how many parsing errors the Kotlin analyzer encounters during analysis.
Tracking files that fail during parsing should help us understand parser robustness and prioritize follow-up improvements.
This follows the same goal and pattern as `SONARIAC-2889: Add telemetry for IaC parse failures per language` adapted to SonarKotlin.

# WHAT
Add telemetry for Kotlin parse failures.

### Definition of Done:
- Files that fail during parsing are counted per analyzed project.
- Telemetry is emitted through the Kotlin analyzer telemetry flow.
- The implementation follows the same overall pattern as `SONARIAC-2889: Add telemetry for IaC parse failures per language` where relevant.
- Tests cover the aggregation/reporting path and at least representative parsing-failure cases.
- Existing telemetry behavior remains unchanged.

# HOW
Suggested implementation direction:
- Reuse existing SonarKotlin telemetry infrastructure instead of emitting telemetry ad hoc from multiple places.
- Track files that fail during parsing and aggregate a final count at project level.
- Use `SONARIAC-2889: Add telemetry for IaC parse failures per language` as the reference for scope and intent, adapted to Kotlin-specific architecture.

### DIFF
--- a/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
+++ b/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
@@ -25,4 +25,5 @@ class TelemetryData {
     var hasAndroidImports = false
     var surefireClassesImported = 0
     var surefireClassesFailed = 0
+    var parseFailuresCounter = 0
 }
--- a/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
+++ b/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
@@ -27,8 +27,10 @@ class TelemetryData {
     var surefireClassesFailed = 0
     // .kt + .kts counters
     var filesProcessed = 0
+    var readFailures = 0
     var parseFailures = 0
     // .kts only counters
     var scriptsProcessed = 0
+    var scriptReadFailures = 0
     var scriptParseFailures = 0
 }
--- a/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
+++ b/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
@@ -25,5 +25,8 @@ class TelemetryData {
     var hasAndroidImports = false
     var surefireClassesImported = 0
     var surefireClassesFailed = 0
+    var filesAnalyzedCounter = 0
     var parseFailuresCounter = 0
+    var scriptsAnalyzedCounter = 0
+    var scriptParseFailuresCounter = 0
 }
--- a/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
+++ b/sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
@@ -25,8 +25,10 @@ class TelemetryData {
     var hasAndroidImports = false
     var surefireClassesImported = 0
     var surefireClassesFailed = 0
+    // .kt + .kts counters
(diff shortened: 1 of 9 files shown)


sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
```diff
@@ -25,4 +25,5 @@ class TelemetryData {
     var hasAndroidImports = false
     var surefireClassesImported = 0
     var surefireClassesFailed = 0
+    var parseFailuresCounter = 0
```
petertrr: Do we have somewhere a total number of successfully parsed files? I don't see it on the Kotlin-specific table, but maybe in one of the general ones?
Anyway, I think it would be valuable to have separate counters for success and failure (to be able to estimate ratio), and I'd also suggest to have separate count for `.kt` and `.kts` files.

petertrr (changes requested):
Implementation LGTM. Please see my comment; I think we can make this telemetry more informative

sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
M1ke-Gru: OK, I will do that. It also might make sense to track successes on the sensors I previously implemented the parse failure tracking on, but I am not sure.

sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
```diff
@@ -25,4 +25,10 @@ class TelemetryData {
     var hasAndroidImports = false
     var surefireClassesImported = 0
     var surefireClassesFailed = 0
+    // .kt + .kts counters
+    var filesAnalyzedCounter = 0
+    var parseFailuresCounter = 0
+    // .kts only counters
+    var scriptsAnalyzedCounter = 0
+    var scriptParseFailuresCounter = 0
```
petertrr: Nitpick: the `Counter` suffix is misaligned with other fields; they should either all have it, or none. I think, from the way these fields are used, it's clear that they are counters, so I'd suggest removing it

sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
```diff
@@ -25,4 +25,10 @@ class TelemetryData {
     var hasAndroidImports = false
     var surefireClassesImported = 0
     var surefireClassesFailed = 0
+    // .kt + .kts counters
+    var filesAnalyzedCounter = 0
```
petertrr: IMO "analyzed" sounds like "successfully analyzed" (or at least it sounds so in my head :) ). But with the current implementation, it's more like "matched" or "processed" - all files that the analyzer receives from the scanner, which are counted before any attempt to read or parse their content. Which is still a valuable metric, but with the current implementation `analyzed != processed - parseError`, because the branch where IOException is handled is not counted (see also the comment from Gitar).

petertrr (changes requested):

sonar-kotlin-metrics/src/main/java/org/sonarsource/kotlin/metrics/TelemetryData.kt
M1ke-Gru: As far as I understand, this telemetry needs to only determine parsing failures, not read failures. Therefore I chose to ignore the Gitar bot comment as it sadly has no direct response option. If you think this is less valuable or that I should add another metric to cover the read error cases, I can do that

petertrr (approved):
LGTM

PR #734, opened 2026-06-10
SONARKT-770 S2245: Add security-context heuristic to reduce false positives

Luqmansonar:
## Summary

Ports the DART-276 security-context heuristic to Kotlin for S2245 (`PseudoRandomCheck`). Pseudorandom number generator usages are now reported only when the surrounding code is plausibly security-relevant.

The heuristic has two parts:

- **Part 1 — file-level crypto import check.** If the compilation unit imports any class under one of these prefixes, all PRNG calls in the file are reported (matches ticket spec):
  - `java.security.`, `javax.crypto.`, `org.springframework.security.`, `org.bouncycastle.`, `io.jsonwebtoken.`, `com.auth0.jwt.`, `at.favre.lib.crypto.`
- **Part 2 — scope-level keyword scan.** If no crypto import is present, walk the parent chain from the PRNG call to find the closest enclosing `KtNamedFunction`; fall back to the closest enclosing `KtClassOrObject` for property initializers and init blocks; finally fall back to the `KtFile` for top-level property initializers. Collect every IDENTIFIER leaf token in that scope, tokenize it (split on `_`, then on camelCase / PascalCase boundaries; all-uppercase parts stay whole), lowercase each token, and intersect with the 32-keyword set from the ticket. Any hit → flag.

## Decisions (confirmed with reviewer)

- **Heuristic applies to all 7 currently-flagged PRNG sources** (`java.util.Random`, `JVMRandom`, `Math.random()`, `kotlin.random.Random`, `ThreadLocalRandom`, `RandomUtils`, `RandomStringUtils`) — not only the two literally listed in the ticket. The ticket's motivation (false positives in non-security code) applies equally to all seven. Consistent with the choice made for the parallel Java port (SONARJAVA-6440).
- **Scope = enclosing named function; falls back to enclosing class/object; finally falls back to the file.** Mirrors Dart's `declarationScope` semantics across Kotlin's broader range of declaration shapes (top-level functions, top-level properties, object declarations, etc.).
- **Tokenization matches the Dart reference exactly** (`_splitIntoWords` from `sonar-dart`). One faithful-port quirk worth knowing: the keyword `randombytes` is all-lowercase, so `randomBytes` (camelCase) which tokenizes to `[random, bytes]` does NOT match — only literal `randombytes` or `RANDOMBYTES` does. Same behavior as Dart and as the Java port; documented in the test sample.
- **Kotlin ticket spec differs from Java ticket spec** and both are honoured as written:
  - Kotlin has **7** crypto-import prefixes (no `de.mkammerer.argon2.` — Java has 8).
  - Kotlin has **32** security keywords (Java has 29). Kotlin adds: `hash`, `key`, `symmetric`.
- **Rule HTML/JSON metadata was NOT modified** — the ticket does not request a doc change. Happy to update separately if desired.

## Test plan

- [x] Existing `PseudoRandomCheckSample.kt` assertions still pass. The existing file contains `secure()` / `secureStrong()` identifiers (Compliant cases at the bottom) which themselves satisfy the keyword scan, so all current `// Noncompliant` markers remain valid without modifying the file.
- [x] New `PseudoRandomCheckSampleNoContext.kt`: neutral file, all calls Compliant (`verifyNoIssue`).
- [x] New `PseudoRandomCheckSampleSecurityKeywords.kt`: camelCase / snake_case / UPPER / method-name / parameter keyword cases + Kotlin-new keywords (`hash`, `key`, `symmetric`); negative cases (`unrelatedScope`, `splitRandomBytes`) do not fire.
- [x] New `PseudoRandomCheckSampleCryptoImport.kt`: `org.springframework.security.*` import → all calls flagged via Part 1.
- [x] New `PseudoRandomCheckSampleTopLevel.kt`: top-level property initializers fall back to file scope; class-scope fallback for class-with-keyword cases (`TokenGenerator`, `Holder` with `secret`).
- [x] `./gradlew :sonar-kotlin-checks:test --tests "org.sonarsource.kotlin.checks.PseudoRandomCheckTest"` → 5 tests, 0 failures.


its/ruling/src/test/resources/expected/kotlin/okio/kotlin-S2245.json
petertrr, line 1: You can remove an empty file from expected results

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
```diff
@@ -56,6 +64,27 @@ private val RANDOM_CONSTRUCTORS_MATCHER = FunMatcher(matchConstructor = true) {
     )
 }
 
+// SONARKT-770: ported from DART-276 security-context heuristic.
```
petertrr: Claude Code tends to leave a lot of comments referencing the way it was working, and they are often too tightly connected with specific prompts. IMO this information is redundant in the code:
* ticket number can be looked up from git history
* connection to Dart is an implementation detail. People who will work with this code in the future do not have to be aware of sonar-dart, let alone the DART Jira project

Could you please clean up these comments?

its/ruling/src/test/resources/expected/kotlin/intellij-rust/kotlin-S2245.json
petertrr, line 1: Same here, an empty file

kotlin-checks-test-sources/src/main/kotlin/checks/PseudoRandomCheckSampleCryptoImport.kt
```diff
@@ -0,0 +1,19 @@
+package checks
+
+import java.util.Random
+import org.springframework.security.crypto.encrypt.Encryptors
+
+// SONARKT-770 Part 1: file imports `org.springframework.security.*` -> all PRNG calls flagged
```
petertrr:
```suggestion
// file imports `org.springframework.security.*` -> all PRNG calls flagged
```

kotlin-checks-test-sources/src/main/kotlin/checks/PseudoRandomCheckSampleSecurityKeywords.kt
```diff
@@ -0,0 +1,78 @@
...
+        return RandomStringUtils.random(1) // Noncompliant
+    }
+
+    // --- Whole-identifier match (`password`). ---
+    fun wholeIdentifierMatch() {
+        val password = "x"
+        val r = Random() // Noncompliant
+    }
+
+    // --- Keyword "key" (new in Kotlin spec, absent from Java). ---
```
petertrr: > new in Kotlin spec, absent from Java

What does it mean?

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
```diff
@@ -65,22 +94,109 @@ class PseudoRandomCheck : CallAbstractCheck() {
...
+        val scope = findDeclarationScope(callExpression) ?: return false
+        return scopeCache.computeIfAbsent(scope, ::computeScopeHasSecurityKeyword)
     }
 }
 
+private fun computeHasCryptoImport(file: KtFile): Boolean =
+    file.importDirectives.any { directive ->
+        val fqn = directive.importedFqName?.asString() ?: return@any false
+        CRYPTO_IMPORT_PREFIXES.any { fqn.startsWith(it) }
+    }
```
petertrr: Nitpick: IMO this is more readable
```suggestion
    file.importDirectives.mapNotNull { directive ->
        directive.importedFqName?.asString()
    }.any { fqn ->
        CRYPTO_IMPORT_PREFIXES.any { fqn.startsWith(it) }
    }
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
```diff
@@ -65,22 +94,109 @@ class PseudoRandomCheck : CallAbstractCheck() {
...
+    collectIdentifierNames(scope).any { name ->
+        tokenizeIdentifier(name).any { it in SECURITY_KEYWORDS }
+    }
+
 private fun KtCallExpression.isChainedMethodInvocation() = parent?.let {
     it is KtDotQualifiedExpression && it.receiverExpression is KtDotQualifiedExpression &&
         (it.receiverExpression as KtDotQualifiedExpression).selectorExpression is KtCallExpression
 } ?: false
+
+// Mirrors Dart `declarationScope`: closest enclosing named function for local code; falls back to
```
petertrr:
```suggestion
// Detects declaration scope: closest enclosing named function for local code; falls back to
```

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
```diff
@@ -65,22 +94,109 @@ class PseudoRandomCheck : CallAbstractCheck() {
...
+// the enclosing class/object for property initializers and init blocks; falls back to the KtFile
+// for top-level property initializers.
+private fun findDeclarationScope(element: PsiElement): PsiElement? {
+    var current: PsiElement? = element.parent
+    while (current != null) {
```
petertrr: Can `element.parents.firstOrNull { ... }` be used instead of a loop?

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
```diff
@@ -65,22 +94,109 @@ class PseudoRandomCheck : CallAbstractCheck() {
...
+    PsiTreeUtil.processElements(scope) { element ->
+        if (element is LeafPsiElement && element.elementType == KtTokens.IDENTIFIER) {
+            names.add(element.text)
+        }
+        true
+    }
+    return names
+}
+
+// Mirrors Dart `_splitIntoWords`: split on `_` first; for each non-empty part, either keep it as
```
petertrr:
```suggestion
// Split on `_` first; for each non-empty part, either keep it as
```

petertrr (changes requested):

its/ruling/src/test/resources/expected/kotlin/okio/kotlin-S2245.json
Luqmansonar: Done in b6fc1a43 - deleted.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
Luqmansonar: Cleaned up across the file in b6fc1a43. Removed ticket references and Dart-implementation comments from `PseudoRandomCheck.kt` and the test samples; remaining comments describe what the code does.

its/ruling/src/test/resources/expected/kotlin/intellij-rust/kotlin-S2245.json
Luqmansonar: Done in b6fc1a43 - deleted.

kotlin-checks-test-sources/src/main/kotlin/checks/PseudoRandomCheckSampleCryptoImport.kt
Luqmansonar: Applied in b6fc1a43.

kotlin-checks-test-sources/src/main/kotlin/checks/PseudoRandomCheckSampleSecurityKeywords.kt
Luqmansonar: Fair point - the Java cross-reference is noise for someone reading sonar-kotlin. Reworded in b6fc1a43 to just say what each test asserts (e.g. `// --- Keyword "key" matches via parameter/variable name. ---`).

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
Luqmansonar: Applied in b6fc1a43.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
Luqmansonar: Applied in b6fc1a43.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
Luqmansonar: Yes - replaced the loop with `element.parents(withSelf = false).firstOrNull { it is KtNamedFunction || it is KtClassOrObject || it is KtFile }`. b6fc1a43.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/PseudoRandomCheck.kt
Luqmansonar: Applied in b6fc1a43.

petertrr (approved):

PR #735, opened 2026-06-11
SONARKT-771 Should not log InvalidFirElementTypeException as error

petertrr:
Part of 

----
## Summary by Gitar

- **Error handling:**
  - Overrode `KtTypeReference.type` in `SonarKaSession` to catch exceptions during type resolution.
  - Returns `KaErrorType` instead of logging `InvalidFirElementTypeException` as an error.
  - Changed `SonarLogger.error` to use `LOG.debug` instead of printing directly to `System.err`.
- **Testing:**
  - Added a regression test in `SonarKaSessionTest` to verify `KaErrorType` is returned for invalid type references.

### DIFF
--- a/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/visiting/KotlinFileVisitor.kt
+++ b/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/visiting/KotlinFileVisitor.kt
@@ -23,12 +23,8 @@ import org.jetbrains.kotlin.analysis.api.analyze
 import org.jetbrains.kotlin.analysis.api.components.KaCompilationResult
 import org.jetbrains.kotlin.analysis.api.components.KaCompilerTarget
 import org.jetbrains.kotlin.analysis.api.diagnostics.KaDiagnostic
-import org.jetbrains.kotlin.analysis.api.types.KaType
 import org.jetbrains.kotlin.config.CompilerConfiguration
-import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.psi.KtFile
-import org.jetbrains.kotlin.psi.KtTypeReference
-import org.slf4j.LoggerFactory
 import org.sonarsource.kotlin.api.checks.InputFileContext
 import org.sonarsource.kotlin.api.frontend.KotlinFileContext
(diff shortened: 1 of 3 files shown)


sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/visiting/KotlinFileVisitor.kt
```diff
@@ -45,6 +46,27 @@ internal var kaSession: KaSession? = null
...
+        get() {
+            return try {
+                with(originalKaSession) {
+                    this@type.type
+                }
+            } catch (e: Exception) {
+                LOG.debug("Unexpected result during type resolution", e)
+                buildClassType(ClassId.fromString("<error>"))
+            }
+        }
```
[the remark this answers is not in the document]
petertrr: We don't handle these exceptions anywhere, so it doesn't matter.

Godin (approved):

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/visiting/KotlinFileVisitor.kt
```diff
@@ -56,6 +58,11 @@ internal class SonarKaSession(
         allowedErrorFilter: (KaDiagnostic) -> Boolean,
     ): KaCompilationResult =
         throw UnsupportedOperationException()
+
+    companion object {
+        @JvmStatic
+        internal val LOG = LoggerFactory.getLogger(SonarKaSession::class.java)
+    }
```
Godin: @petertrr maybe move it into `SonarLogger` ?

Godin (changes requested):

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/visiting/KotlinFileVisitor.kt
petertrr: My motivation was that having a logger named "SonarKaSession" will be more meaningful than "SonarLogger"  for the case of type resolution

Godin (approved):

PR #739, opened 2026-06-15
SONARKT-777 ToggleLockBranch: Add additional-message

pavel-mikula-sonarsource:
This new parameter will be appended to the Slack message. It removes the need to go to Slack and explain why it's being locked

### DIFF
--- a/.github/workflows/ToggleLockBranch.yml
+++ b/.github/workflows/ToggleLockBranch.yml
@@ -13,6 +13,10 @@ on:
         description: "Branch to to toggle lock on"
         required: true
         default: "master"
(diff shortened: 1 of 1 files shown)


.github/workflows/ToggleLockBranch.yml
```diff
@@ -13,6 +13,10 @@ on:
         description: "Branch to to toggle lock on"
         required: true
         default: "master"
+    inputs:
```
pavel-mikula-sonarsource:
```suggestion
```

jonas-wielage-sonarsource (approved):

PR #745, opened 2026-07-01
SONARKT-782 S2068 and S6418: make use of common secret exclusion filter

matthew-reid-sonarsource:
## What

Adopts the shared `SecretClassifier` API from **sonar-analyzer-commons** (`org.sonarsource.analyzer.commons.appsec.SecretClassifier`, introduced in [sonar-analyzer-commons#411](https://github.com/SonarSource/sonar-analyzer-commons/pull/411) / SONARIAC-3020) so that **S2068** (`HardcodedCredentialsCheck`) and **S6418** (`HardcodedSecretsCheck`) only raise an issue on a value `v` when `SecretClassifier.isKnownNonSecret(v)` returns `false`.

This skips values that look like fake credentials, placeholders, variable references, encrypted markers, external-store references, or recognizable structured values — reducing false positives. Part of phase 2 of [APPSEC-3451](https://sonarsource.atlassian.net/browse/APPSEC-3451), mirroring the sonar-iac-enterprise adoption in [sonar-iac-enterprise#1092](https://github.com/SonarSource/sonar-iac-enterprise/pull/1092).

## How

Both rules funnel their per-value decision through `AbstractHardcodedVisitor.isSensitiveStringLiteral(value)`, which covers both the `word=value` literal path and the variable-assignment path. The classifier is gated there once:

```kotlin
open fun isSensitiveStringLiteral(value: String): Boolean {
    return value.isNotEmpty() && !SecretClassifier.isKnownNonSecret(value)
}
```

and composed into the S6418 override (via `super.isSensitiveStringLiteral(value)`) ahead of the existing entropy / length / human-language checks.

The `*Sample.kt` test files are updated: `// Noncompliant` cases whose values are now correctly recognized as non-secrets (e.g. `"xxx"`, `"login=a&password=xxx"`, `"abcdefghijklmnopqrs"` — the last matches the `abcd` fake-word pattern) are switched to realistic high-entropy values so the intended issue still raises, and a few explicit `// Compliant` cases demonstrate the new exclusions.

Scope note: matching the IaC PR, the S2068 hard-coded-URL-with-credentials path (`http://user:pass@host`) is intentionally left untouched — the classifier gates the value-sensitivity predicate only.

## Testing

`:sonar-kotlin-checks:test` passes locally (full module, including `HardcodedCredentialsCheckTest` and `HardcodedSecretsCheckTest`)

[APPSEC-3451]: https://sonarsource.atlassian.net/browse/APPSEC-3451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/HardcodedSecretsCheck.kt
```diff
@@ -61,7 +61,7 @@ class HardcodedSecretsCheck : AbstractHardcodedVisitor() {
     }
 
     override fun isSensitiveStringLiteral(value: String): Boolean {
-        return value.isNotEmpty()
+        return super.isSensitiveStringLiteral(value)
                 && value.length >= MINIMUM_CREDENTIAL_LENGTH
```
teemu-rytilahti-sonarsource: The minimum length is also checked by the SecretsClassifier.

teemu-rytilahti-sonarsource (approved):
LGTM with a minor nit on the length limit.

PR #746, opened 2026-07-01
SONARKT-784 Detect test files with a reusable TestFileClassifier and suppress S2068/S6418 on them

teemu-rytilahti-sonarsource:
## What

Introduces `TestFileClassifier` — a reusable, registration-based classifier for
detecting test/fixture files — and uses it to stop the hardcoded-secret rules
S2068 and S6418 from raising on test files.

When a project does not set `sonar.tests`, the scanner labels test files as MAIN,
so these rules fire on test fixtures containing placeholder credentials — a
common source of false positives.

## How

- `TestFileClassifier` (Java, in `sonar-kotlin-api`): `of(configuration, globs…)`
  registers a language's test-file path globs and evaluates the gate once —
  `sonar.tests` / `sonar.test.inclusions` / `sonar.test.exclusions`, plus a
  generic opt-out `sonar.testFileHeuristic.disabled`. When no globs are
  registered, it falls back to a generic set of test directories (`**/test/**`,
  `**/tests/**`, `**/__tests__/**`). The per-file `looksLikeTestFile(InputFile)`
  emits a one-time "set sonar.tests" warning on first hit. An
  `of(configuration, detector, globs…)` overload takes a caller-supplied
  detector (`Predicate<Context>`) for checks richer than path matching (e.g. a
  parsed tree that imports a test framework); `Context` is a caller-defined
  per-file input the detector downcasts to. Written in Java so it can later be
  lifted verbatim to `sonar-analyzer-commons`.
- The Kotlin sensor registers the scope and computes a per-file `isTestFile`
  (platform `InputFile.Type.TEST` or the heuristic), exposed on
  `InputFileContext`; S2068 and S6418 consult it and skip test files.
- Classification affects rule execution only, never metrics.
- Kotlin patterns: `**/test/**`, `**/tests/**`, `**/*Test.kt`, `**/*Tests.kt`,
  `**/*Spec.kt`, `**/*IT.kt`.

### DIFF
--- a/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedCredentialsInTestFileSample.kt
+++ b/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedCredentialsInTestFileSample.kt
@@ -1,0 +1,11 @@
+package checks
+
+// No "Noncompliant" markers: these credentials would be reported in main code,
+// but the check is expected to stay silent when the file is classified as a test file.
+internal class HardcodedCredentialsInTestFileSample {
+    fun f() {
+        val params = "user=admin&password=Password123"
+        var passwd = "xxxx"
+        passwd = "yyyy"
+    }
+}
--- a/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedCredentialsInTestFileSample.kt
+++ b/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedCredentialsInTestFileSample.kt
@@ -1,11 +1,12 @@
 package checks
 
-// No "Noncompliant" markers: these credentials would be reported in main code,
-// but the check is expected to stay silent when the file is classified as a test file.
+// These values would trigger S2068 in main code but are suppressed when the file is a test file.
 internal class HardcodedCredentialsInTestFileSample {
     fun f() {
-        val params = "user=admin&password=Password123"
-        var passwd = "xxxx"
-        passwd = "yyyy"
+        println("login=a&password=Rb7kZpQ2")
+        var passwd = "Rb7kZpQ2"
+        println(passwd)
+        passwd = "Rb7kZpQ2"
+        println(passwd)
     }
 }
--- a/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedSecretsInTestFileSample.kt
+++ b/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedSecretsInTestFileSample.kt
@@ -1,0 +1,10 @@
+package checks
+
+// No "Noncompliant" markers: these secrets would be reported in main code,
+// but the check is expected to stay silent when the file is classified as a test file.
+internal class HardcodedSecretsInTestFileSample {
+    fun f() {
+        val mySecret = "abcdefghijklmnopqrs"
+        val params = "login=a&secret=abcdefghijklmnopqrs"
+    }
+}
--- a/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedSecretsInTestFileSample.kt
+++ b/kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedSecretsInTestFileSample.kt
@@ -1,10 +1,10 @@
 package checks
 
-// No "Noncompliant" markers: these secrets would be reported in main code,
-// but the check is expected to stay silent when the file is classified as a test file.
+// These values would trigger S6418 in main code but are suppressed when the file is a test file.
 internal class HardcodedSecretsInTestFileSample {
     fun f() {
-        val mySecret = "abcdefghijklmnopqrs"
-        val params = "login=a&secret=abcdefghijklmnopqrs"
+        val mySecret = "Hj4pZ9wLdN2sKq7VtXy"
+        println("login=a&secret=Hj4pZ9wLdN2sKq7VtXy")
+        println(mySecret)
     }
 }
--- a/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
+++ b/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
@@ -109,8 +109,6 @@ public final class TestFileClassifier {
 
   private static boolean isTestSourceConfigured(Configuration config) {
     return isSet(config, "sonar.tests")
-      || isSet(config, "sonar.test.inclusions")
-      || isSet(config, "sonar.test.exclusions")
       || config.getBoolean(HEURISTIC_DISABLED_KEY).orElse(false);
   }
 
--- a/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
+++ b/sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
@@ -1,0 +1,125 @@
+/*
+ * SonarSource Kotlin
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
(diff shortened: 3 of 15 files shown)


daniel-teuchert-sonarsource (changes requested):
Overall LGTM!
Can we also add a default filter logic (e.g. if the path contains `test` or the file begins with `test`) for analyzers that don't want to define their own filtering logic?


The ticket SONARKT-784 is about enabling test file filtering for S2068 and S6418 but I think this PR enables it for all rules. If I did not misunderstand this we should either adjust the ticket or this PR.
If you want to enable test file filtering, can you check if this interface would also work for individual rules just to make sure it will work for the other analyzers?

teemu-rytilahti-sonarsource:
> Can we also add a default filter logic (e.g. if the path contains `test` or the file begins with `test`) for analyzers that don't want to define their own filtering logic?

There is now a (directory-only) fallback to keep it strict and only exact directory names as different ecosystems have so many different variations how this can be done. I can extend it if you think it makes sense, my thinking was to keep the default as minimal and precise as possible to avoid FNs. As Kotlin will already define its own globs, this does not affect it & we can adjust when we move the class out to commons. Let me know what do you think!

> The ticket [SONARKT-784](https://sonarsource.atlassian.net/browse/SONARKT-784) is about enabling test file filtering for S2068 and S6418 but I think this PR enables it for all rules. If I did not misunderstand this we should either adjust the ticket or this PR. If you want to enable test file filtering, can you check if this interface would also work for individual rules just to make sure it will work for the other analyzers?

I rechecked and the only place the `isTestFile` is checked is in `AbstractHardcodedVisitor` which is the base class for S2068 and S6418.

[SONARKT-784]: https://sonarsource.atlassian.net/browse/SONARKT-784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

daniel-teuchert-sonarsource (commented):
Thanks for adding the default filter. I think that is a good approach!

Also, thank you for double checking and sorry about the confusion as this change already only applies to the two secret rules.

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
```diff
@@ -0,0 +1,130 @@
...
+ * test sources.
+ *
+ * <p>Registered once with the project {@link Configuration} (which gates the heuristic) plus path
+ * globs and, optionally, a {@code detector} for richer checks that reads a caller-supplied
+ * {@link Context}. Per file, {@link #looksLikeTestFile(InputFile)} returns the classification.
+ */
+public final class TestFileClassifier {
+
+  /** Generic opt-out: disables the test-file heuristic for every analyzer that uses this classifier. */
+  public static final String HEURISTIC_DISABLED_KEY = "sonar.testFileHeuristic.disabled";
```
teemu-rytilahti-sonarsource, line 42: We need to decide where we want to store this (under `sonar.test.*`?) and document it.

kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedCredentialsInTestFileSample.kt
```diff
@@ -0,0 +1,11 @@
+package checks
+
+// No "Noncompliant" markers: these credentials would be reported in main code,
+// but the check is expected to stay silent when the file is classified as a test file.
+internal class HardcodedCredentialsInTestFileSample {
+    fun f() {
+        val params = "user=admin&password=Password123"
+        var passwd = "xxxx"
+        passwd = "yyyy"
```
daniel-teuchert-sonarsource: I think those dummy passwords would not be detected even in main code.

kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedSecretsInTestFileSample.kt
```diff
@@ -0,0 +1,10 @@
+package checks
+
+// No "Noncompliant" markers: these secrets would be reported in main code,
+// but the check is expected to stay silent when the file is classified as a test file.
+internal class HardcodedSecretsInTestFileSample {
+    fun f() {
+        val mySecret = "abcdefghijklmnopqrs"
+        val params = "login=a&secret=abcdefghijklmnopqrs"
+    }
```
daniel-teuchert-sonarsource: Same as above. I think those should be more real looking secrets to ensure that the test file filter suppresses the issue.

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
daniel-teuchert-sonarsource: Good question...
Do we have any guidelines?
I am okay with `sonar.testFileHeuristic.disabled`.

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(TestFileClassifier.class);
+
+  private static final String HEURISTIC_APPLIED_WARNING =
+    "Test files were detected using a path heuristic because \"sonar.tests\" is not set. To improve the " +
+      "analysis accuracy, it is recommended to configure it, e.g.: \"sonar.tests=src/test\".";
```
daniel-teuchert-sonarsource, line 48: Should this warning also mention how to disable the test file heuristic?
For me it is also okay to not mention it. Just an idea.

daniel-teuchert-sonarsource (approved):
LGTM!
I left a few comments about the tests and the logged warning but those are all up to you.

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
teemu-rytilahti-sonarsource: Both are fine for me, too. I think it residing in a hierarchy under `sonar.test.` feels more logical to me, but I could not find any direct guideline on the property ownerships.

In the documentation, [parameters not settable in the UI docs](https://docs.sonarsource.com/sonarqube-server/2026.1/analyzing-source-code/analysis-parameters/parameters-not-settable-in-ui) describe some of the available properties, only a single documented one uses the `x.y.disabled` hierarchy as the python implementation.

There is separate documentation for [the test exclusions](https://docs.sonarsource.com/sonarqube-cloud/managing-your-projects/project-analysis/setting-analysis-scope/excluding-files-based-on-patterns), but unsure if that's the correct location for documenting this new flag.

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/checks/TestFileClassifier.java
teemu-rytilahti-sonarsource: I hope that this is only a temporary message, and we can show a proper warning in the UI like it is done for python. I'd say let's keep it simple, and document it elsewhere as necessary.

daniel-teuchert-sonarsource: Agreed

kotlin-checks-test-sources/src/main/kotlin/checks/HardcodedCredentialsInTestFileSample.kt
teemu-rytilahti-sonarsource: Fixed.

PR #748, opened 2026-07-09
SONARKT-788 Document that bumping kotlinVersion requires a SonarSecurity update

malte-skoruppa-sonarsource:
Adds a note next to `kotlinVersion` in `gradle.properties` documenting the cross-repo coupling with SonarSecurity: SonarSecurity's Kotlin frontend is compiled against the exact Kotlin compiler version that SonarKotlin provides at runtime, so any bump here must be mirrored in SonarSecurity (`kotlin.version` / `sonarkotlin.version`, plus any required Kotlin Analysis API migration).

Context: the recent Kotlin 2.4.0 bump required a coordinated update and Analysis API migration in SonarSecurity (SONARSEC-8626). This comment makes the coupling discoverable at the point of change.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,6 +2,10 @@ group=org.sonarsource.kotlin
 version=3.8-SNAPSHOT
 description=Code Analyzer for Kotlin
 projectTitle=Kotlin
+# IMPORTANT: When bumping kotlinVersion, the Kotlin compiler version MUST also be updated in
+# SonarSecurity, which compiles its Kotlin frontend against this exact version (provided at
+# runtime by SonarKotlin). Update SonarSecurity's `kotlin.version` and `sonarkotlin.version`
+# properties together, plus any required migration of its Kotlin frontend to the new API.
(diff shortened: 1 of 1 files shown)


gradle.properties
felix-pauck-sonarsource, line 1: Like it!

We could add links or contact information to help performing this task, however, since these information sometimes change, I think it is fine to also not put them here.

Continue as you like! 😊

felix-pauck-sonarsource (approved):
LGTM! 👍

gradle.properties
malte-skoruppa-sonarsource: Yes, I thought about it as well. And, as you said, this can change. I am sure anyone updating the Kotlin compiler version in SonarKotlin is smart enough to figure out how to update the Kotlin compiler + SonarKotlin dependencies in SonarSecurity, and also who to reach out to in case of problems, so let's keep it as it is. :slightly_smiling_face:

yasen-pavlov-sonarsource (approved):
🙏

PR #749, opened 2026-07-13
SONARKT-790 Plugin version telemetry

M1ke-Gru:
----
## Summary by Gitar

- **Telemetry enhancement:**
  - Added `kotlin.pluginVersion` to telemetry metrics for tracking plugin versions.
  - Refactored `KotlinProjectSensor` to include helper methods for consistent logging and property registration.
- **Build configuration:**
  - Integrated `org.sonarsource.cloud-native.project-version-provisioning` plugin.
  - Updated `gradle/verification-metadata.xml` with new dependency checksums.
  - Updated `build-logic/common` submodule reference.
- **Repository housekeeping:**
  - Added `.kotlin/` to `.gitignore` and removed stale error logs from the directory.


sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinProjectSensor.kt
```diff
@@ -16,34 +16,70 @@
...
+                KotlinProjectSensor::class.java.classLoader.getResourceAsStream(resourceName)?.use { stream ->
+                    val props = Properties()
+                    props.load(stream)
+                    return props.getProperty("plugin.version", "unknown")
+                }
+            } catch (e: IOException) {
+                // fall through to fallback
+            }
+            return "unknown"
+        }
```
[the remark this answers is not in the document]
M1ke-Gru: Fixed in 2c7fa6cd: `resolvePluginVersion()` now falls back to `unknown` if the resolved value still contains an unsubstituted `${` placeholder, with a regression test.

gradle/verification-metadata.xml
```diff
@@ -115,6 +115,11 @@
             <sha256 value="d9d5e5ce724a4071c3b6d96433744f3b69a27c66e0f2d7ed372b7a1df612c0a1" origin="Verified"/>
          </artifact>
       </component>
+      <component group="com.diffplug.spotless" name="spotless-lib" version="4.6.2">
+         <artifact name="spotless-lib-4.6.2.jar">
+            <sha256 value="cdb34b6492938b076b1c84412e9c3cfafbda5d7a06956b5b910e44f82b3e6779" origin="Generated by Gradle"/>
```
petertrr: Please ensure that there are no unexpected dependencies here and change `Generated by Gradle` -> `Verified`

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinProjectSensor.kt
```diff
@@ -16,34 +16,71 @@
...
+    }
+
     override fun describe(descriptor: SensorDescriptor) {
         descriptor.onlyOnLanguage(KOTLIN_LANGUAGE_KEY).name("KotlinProjectSensor")
     }
 
+    fun addAndLogTelemetryProperty(context: SensorContext, propertyName: String, propertyValue: String) {
+        context.addTelemetryProperty(propertyName, propertyValue);
+        LOG.debug("TELEMETRY: $propertyName=$propertyValue");
```
petertrr: I'm a bit out of the loop here: is this to align behavior between all analyzers? Or what's the motivation to also log all measures?

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinProjectSensor.kt
```diff
@@ -16,34 +16,71 @@
...
+    }
+
+    fun fileProcessingTelemetry(context: SensorContext) {
+        // files - .kt and .kts counters
+        addAndLogTelemetryProperty(context, "kotlin.files.processed", telemetryData.filesProcessed.toString())
```
petertrr: You could make this code less verbose by wrapping everything into `with (telemetryData)`, and making `addAndLogTelemetryProperty` accept `Any` and call `.toString()` already inside 😉 But that's opinionated; feel free to dismiss this suggestion

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinProjectSensor.kt
```diff
@@ -16,34 +16,71 @@
...
+        private const val PLUGIN_VERSION_RESOURCE = "org/sonar/plugins/kotlin/pluginVersion.properties"
+
+        fun resolvePluginVersion(): String = resolvePluginVersion(PLUGIN_VERSION_RESOURCE)
+
+        internal fun resolvePluginVersion(resourceName: String): String {
+            try {
+                KotlinProjectSensor::class.java.classLoader.getResourceAsStream(resourceName)?.use { stream ->
+                    val props = Properties()
+                    props.load(stream)
+                    val version = props.getProperty("plugin.version", "unknown")
```
petertrr: The string `"unknown"` is duplicated three times

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinProjectSensor.kt
```diff
@@ -16,34 +16,71 @@
...
+                    return if (version.contains("\${")) "unknown" else version
+                }
+            } catch (e: IOException) {
```
petertrr: Should this be just `Exception`? For example, `props.load` can throw an `IllegalArgumentException` according to javadoc

petertrr (changes requested):

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinProjectSensor.kt
M1ke-Gru: Yes, it is to align it between analyzers. Currently the easiest way to do telemetry validation is to go to peachee repos github actions and look at the debug logs. It was done this way in IaC and text. This is why I am adding this to other repos as well. Probably should have included it in the ticket description though.

gradle/verification-metadata.xml
```diff
@@ -756,6 +801,11 @@
             <sha256 value="a190159b65e82f424e89ce8d91f40dab0e0ad99d9c735e5c90cd4b82119d6c9a" origin="Verified"/>
          </artifact>
       </component>
+      <component group="com.sonarsource.rule-api" name="rule-api" version="2.20.0.5857">
+         <artifact name="rule-api-2.20.0.5857.jar">
+            <sha256 value="1091f5dbbc6ada5fdeb7b543b56d086b18f2d7cfb5a65cd2058a62d387916ae4" origin="Generated by Gradle"/>
```
petertrr: 👀

petertrr (approved):
LGTM, just one remaining item

PR #751, opened 2026-07-20
SONARKT-391 Do not report S6313 on functions annotated with `@VisibleForTesting`

petertrr:
## Summary
- Rule S6313 flagged non-private suspending functions in a `ViewModel` even when they were annotated with `` `@VisibleForTesting` `` (androidx.annotation.VisibleForTesting), which is a false positive since such functions are intentionally exposed only for tests.
- The check now also excludes functions annotated with `` `@VisibleForTesting` `` (matched by simple annotation name, consistent with how other checks in this codebase recognize annotations such as `` `@Test` ``).

## Test plan
- [x] Added a `` `@VisibleForTesting` `` suspending function case to `ViewModelSuspendingFunctionsCheckSample.kt` and verified it is not reported.
- [x] Added an `androidx.annotation.VisibleForTesting` stub under `kotlin-checks-test-sources` (mirroring existing `androidx.lifecycle.ViewModel` stub) so the sample compiles with semantics.
- [x] `./gradlew :sonar-kotlin-checks:test --tests "org.sonarsource.kotlin.checks.ViewModelSuspendingFunctionsCheckTest"` passes.
- [x] `./gradlew :sonar-kotlin-checks:test` (full module) passes.
- [x] `./gradlew spotlessCheck` passes.

Fixes SONARKT-391.

### DIFF
--- a/kotlin-checks-test-sources/src/main/java/androidx/annotation/VisibleForTesting.java
+++ b/kotlin-checks-test-sources/src/main/java/androidx/annotation/VisibleForTesting.java
@@ -1,4 +1,13 @@
 package androidx.annotation;
 
+// Local stub of androidx.annotation.VisibleForTesting, mirroring its public API,
+// so sample files can compile without pulling in the real androidx dependency.
 public @interface VisibleForTesting {
+
+    int otherwise() default PRIVATE;
+
+    int PROTECTED = 4;
+    int PACKAGE_PRIVATE = 3;
+    int PRIVATE = 2;
+    int NONE = 5;
 }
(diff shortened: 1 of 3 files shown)


kotlin-checks-test-sources/src/main/java/androidx/annotation/VisibleForTesting.java
```diff
@@ -0,0 +1,4 @@
+package androidx.annotation;
+
+public @interface VisibleForTesting {
```
mstachniuk, line 5: This definition requires some comment as it already exist https://developer.android.com/reference/androidx/annotation/VisibleForTesting
I guess you create it locally to not import the whole package?

And the definition should be as much similar to the original definition, for example it can be used in following way: `@VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)` - please addd such examples to tests.

mstachniuk (approved):
LGTM

PR #759, opened 2026-07-29
SONARKT-794 Replace orchestrator with SIT for sonar-kotlin

maksim-grebeniuk-sonarsource:
Part of 

----
## Summary by Gitar

- **Integration tests migration:**
    - Replaced SonarQube Orchestrator with `sonar-scanner-integration-tester` (SIT) across plugin and ruling integration test modules
    - Added `its:sq-integration` module to retain tests requiring a live SonarQube server and orchestrator
- **Build configuration:**
    - Updated dependencies to use SIT `1.2.0.1354` and newer plugin API versions in `settings.gradle.kts`
    - Configured Artifactory repository resolution for internal dependencies in `build.gradle.kts`


settings.gradle.kts
```diff
@@ -81,6 +83,9 @@ dependencyResolutionManagement {
             val sonarlint = version("sonarlint", "10.13.0.79996")
             val sonarqube = version("sonarqube", "25.1.0.102122")
             val scannerEngine = version("scannerEngine", "13.4.1.4007")
+            val sit = version("sit", "1.2.0.1354")
+            val sonarPluginApiSit = version("sonarPluginApiSit", "12.0.0.2960")
```
petertrr: Compatibility matrix for plugin-api shows 13.2 as currently supported version, and other plugins (e.g. sonar-text) already use it. I suggest to also update plugin-api here (or update verification metadata and merge #756 first)

its/sq-integration/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+tasks.integrationTest {
+    dependsOn(":sonar-kotlin-plugin:dist")
+    listOf("keepSonarqubeRunning", "reportAll", "cleanProjects", "buildProjects")
+        .associateWith { System.getProperty(it) }
+        .filter { it.value != null }
+        .forEach { systemProperty(it.key, it.value) }
+    systemProperty("java.awt.headless", "true")
+    // export a classpath containing kotlin standard dependencies, needed by the
+    // kotlin-language-server ruling corpus for sonar.java.libraries
+    systemProperty("gradle.main.compile.classpath", sourceSets.main.get().compileClasspath.asPath)
```
petertrr: Why do we need to provide this property externally if the test is running a normal Gradle build?

settings.gradle.kts
maksim-grebeniuk-sonarsource: Good catch

its/sq-integration/build.gradle.kts
maksim-grebeniuk-sonarsource: this was copy-pasted from `its/ruling/build.gradle.kts`

settings.gradle.kts
petertrr: I meant to reconcile again at single dependency and use 13.2 for the entire plugin :)

its/ruling/build.gradle.kts
```diff
@@ -1,28 +1,43 @@
...
         .associateWith { System.getProperty(it) }
         .filter { it.value != null }
         .forEach { systemProperty(it.key, it.value) }
-    systemProperty("java.awt.headless", "true")
     // export a classpath containing kotlin standard dependencies
     systemProperty("gradle.main.compile.classpath", sourceSets.main.get().compileClasspath.asPath)
+    // Each SIT run leaks engine class loaders; a fresh JVM per class keeps heap bounded.
+    setForkEvery(1)
+    // The ruling scan holds the whole its/sources corpus in memory; run it alone rather than in parallel forks.
+    maxParallelForks = 1
```
petertrr, line 40: We used to run old ruling tests in parallel, and it was manageable for the CI machine. Maybe this can be tweaked for SIT too?

petertrr (approved):
LGTM
Let's roll it back to sequential execution and investigate parallelism separately

PR #760, opened 2026-07-30
SONARKT-801 Contain per-file analysis crashes so one bad file doesn't abort the whole Kotlin scan

matthew-reid-sonarsource:
----
## Summary by Gitar

- **Error Handling:**
    - Catch `StackOverflowError` during file parsing and visitor execution to prevent whole-scan aborts
    - Report analysis errors for crashed files and log short messages without full stack traces
- **Telemetry & Warnings:**
    - Track analysis crashes in `TelemetryData` and report project telemetry properties
    - Post customer-visible `AnalysisWarnings` for files impacted by internal analysis crashes


sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
```diff
@@ -137,6 +147,15 @@ abstract class AbstractKotlinSensorExecuteContext(
...
                 null
+            } catch (e: StackOverflowError) {
+                // A StackOverflowError is a java.lang.Error, not an Exception, so it escapes the catch above.
+                // It can originate in the recursive PSI construction of a deeply-nested file. Contain it here so
+                // the up-front parse of all files does not abort the whole scan. Keep the handler stack-frugal:
+                // log a short message and do NOT pass the throwable to the logger (avoid formatting its huge trace).
+                inputFileContext.reportAnalysisParseError(KOTLIN_REPOSITORY_KEY, it, null)
+                onAnalysisCrash(it)
+                logger.error("Cannot parse '$it': ${e.javaClass.name}")
+                null
```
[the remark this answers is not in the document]
matthew-reid-sonarsource: This is intentional, if I made your suggested change then this would be the only throwable in the parse phase that aborts the scan

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
```diff
@@ -137,6 +148,15 @@ abstract class AbstractKotlinSensorExecuteContext(
                 inputFileContext.reportAnalysisParseError(KOTLIN_REPOSITORY_KEY, it, parseException.position)
                 onReadFailure()
                 null
+            } catch (e: StackOverflowError) {
+                // A StackOverflowError is a java.lang.Error, not an Exception, so it escapes the catch above.
+                // It can originate in the recursive PSI construction of a deeply-nested file. Contain it here so
+                // the up-front parse of all files does not abort the whole scan. Keep the handler stack-frugal:
+                // log a short message and do NOT pass the throwable to the logger (avoid formatting its huge trace).
```
petertrr: Could you change the comment to something shorter? 😅 Maybe:
```suggestion
                // In case the Kotlin compiler crashes with SOE on e.g. recursive types,
                // do not abort the whole scan; log a message without a huge stack trace
```

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
```diff
@@ -184,6 +205,23 @@ abstract class AbstractKotlinSensorExecuteContext(
                 measureDuration(visitorId) {
                     visitor.scan(inputFileContext, tree)
                 }
+            } catch (e: StackOverflowError) {
+                // A StackOverflowError is a java.lang.Error, not an Exception, so it escapes the catch below.
+                // It can originate deep in the Kotlin compiler's type resolver (e.g. an invalid cyclic typealias)
+                // or in the analyzer's own recursive PSI walkers. Contain it here so a single crashing file is
+                // reported and skipped instead of aborting the whole scan. This handler runs at a shallow frame
+                // (the deep frames have already unwound), so it has ample stack headroom -- but keep it frugal:
+                // log a short message and do NOT pass the throwable to the logger (avoid formatting its huge trace).
```
petertrr: Same here

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
```diff
@@ -184,6 +205,23 @@ abstract class AbstractKotlinSensorExecuteContext(
                 measureDuration(visitorId) {
                     visitor.scan(inputFileContext, tree)
                 }
+            } catch (e: StackOverflowError) {
```
petertrr: Nitpick: why is catch for SOE first here, and last in the parsing block?

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
```diff
@@ -184,6 +205,23 @@ abstract class AbstractKotlinSensorExecuteContext(
...
+                // A StackOverflowError is a java.lang.Error, not an Exception, so it escapes the catch below.
+                // It can originate deep in the Kotlin compiler's type resolver (e.g. an invalid cyclic typealias)
+                // or in the analyzer's own recursive PSI walkers. Contain it here so a single crashing file is
```
petertrr: > or in the analyzer's own recursive PSI walkers

IMO this fix is aimed at the compiler's code that we can't control, and we shouldn't document safeguarding our own checks as a purpose

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
```diff
@@ -184,6 +205,23 @@ abstract class AbstractKotlinSensorExecuteContext(
...
+                // reported and skipped instead of aborting the whole scan. This handler runs at a shallow frame
+                // (the deep frames have already unwound), so it has ample stack headroom -- but keep it frugal:
+                // log a short message and do NOT pass the throwable to the logger (avoid formatting its huge trace).
+                val message = "Analysis of '${inputFileContext.inputFile}' with '$visitorId' failed: ${e.javaClass.name}"
```
petertrr: Should we have the same format (`"Cannot analyse..."`)  for both branches? This will make analysis of logs easier IMO. And these recovered crashes can be detected as something like `Cannot analyse.*Error`

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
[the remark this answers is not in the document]
matthew-reid-sonarsource: The asymmetry is intentional

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinPlugin.kt
```diff
@@ -64,6 +65,12 @@ class KotlinPlugin : Plugin, KotlinPluginExtensionsProvider {
 
         context.addExtension(KotlinGradleSensor::class.java)
 
+        if (context.runtime.product == SonarProduct.SONARLINT) {
+            // The scanner provides AnalysisWarnings; SonarLint does not. Register a no-op so the sensors (which
+            // require AnalysisWarnings via their constructor) can be instantiated in the SonarLint container.
+            context.addExtension(NoOpAnalysisWarnings::class.java)
```
petertrr, line 69: Curious: sonar-iac also has a similar `NOOP_ANALYSIS_WARNINGS`, but I don't see conditional registering of beans there. Do you know why it works differently?

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
```diff
@@ -777,6 +886,18 @@ internal class ExceptionThrowingCheck : AbstractCheck() {
     }
 }
 
+@Rule(key = "SOERule1")
+internal class StackOverflowThrowingCheck : AbstractCheck() {
+    override fun visitNamedFunction(function: KtNamedFunction, kfc: KotlinFileContext) {
+        // Deterministically exercise the StackOverflowError containment path (Change 1) without depending on the
```
petertrr: > Change 1

🤔

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
```diff
@@ -64,6 +67,9 @@ private val LOG = LoggerFactory.getLogger(KotlinSensor::class.java)
 internal class KotlinSensorTest : AbstractSensorTest() {
     private val disposable = Disposer.newDisposable()
 
+    // Fresh per test (JUnit5 default PER_METHOD lifecycle) so crash-warning verifications don't bleed across tests.
```
petertrr: Maybe a bit more general
```suggestion
    // Fresh per test (JUnit5 default PER_METHOD lifecycle) so analysis warnings don't leak across test
```

sonar-kotlin-plugin/src/test/java/org/sonarsource/kotlin/plugin/KotlinSensorTest.kt
```diff
@@ -389,6 +395,109 @@ internal class KotlinSensorTest : AbstractSensorTest() {
...
+        context.fileSystem().add(createInputFile("good1.kt", "class A { fun a() = TODO() }"))
+        context.fileSystem().add(createInputFile("good2.kt", "class B { fun b() = TODO() }"))
+        sensor(checkFactory("S1764")).execute(context)
+
+        verify(exactly = 0) { analysisWarnings.addUnique(any()) }
+    }
+
+    @Test
+    fun `a StackOverflowError while building the syntax structure is contained (parse path)`() {
+        // Regression for Change 2 (the SOE catch in the kotlinFiles parse lazy). A StackOverflowError raised inside
```
petertrr: I guess `Change 2 (the SOE catch in the kotlinFiles parse lazy)` references something from the ticket, but IMO comments in code should be self-contained and describe only the actual code

petertrr (changes requested):
Implementation LGTM with one small suggestion about logging format. Other than that: you know I have some opinions about LLM comments 😅 , so feel free to disagree if you think some comments are needed in full length, but could you please fix those where specification details are leaking into code?

sonar-kotlin-api/src/main/java/org/sonarsource/kotlin/api/sensors/AbstractKotlinSensorExecuteContext.kt
matthew-reid-sonarsource: No reason, the order was completely arbitrary, but for consistency's sake I made the Error catch last in both blocks now

sonar-kotlin-plugin/src/main/java/org/sonarsource/kotlin/plugin/KotlinPlugin.kt
matthew-reid-sonarsource: Mostly the different shape. In iac, `NOOP_ANALYSIS_WARNINGS` isn't a bean, it's a static final constant passed in by a fallback constructor (`TerraformProviders(Configuration) -> this(config, NOOP_ANALYSIS_WARNINGS)`). The only bean is `DefaultAnalysisWarningsWrapper`, behind their own `@ScannerSide` interface, registered unconditionally. The annotation filter keeps it out of SonarLint. So there's nothing there to register conditionally.


Here the no-op is a bean, which is what forces the guard: `AnnotationUtils.getAnnotation` walks implemented interfaces and `AnalysisWarnings` is itself `@ScannerSide`, so `NoOpAnalysisWarnings` matches the scanner's filter too. Registered unconditionally it lands in the project container alongside the real `DefaultAnalysisWarnings` from the global container, and PriorityBeanFactory breaks the tie in favor of the child, so the sensors would get the no-op and scanner warnings would silently vanish.

petertrr (approved):
LGTM

PR #762, opened 2026-07-31
SONARKT-800 Surefire importer should not crash when multiple reports map to the same file

petertrr:


sonar-kotlin-surefire/src/main/java/org/sonarsource/kotlin/surefire/KotlinSurefireParser.java
```diff
@@ -116,12 +123,21 @@ private void save(UnitTestIndex index, SensorContext context) {
...
+            telemetryData.setSurefireClassesImported(telemetryData.getSurefireClassesImported() + 1);
+          } else {
+            LOGGER.debug("Measures for {} were already saved, skipping to avoid saving the same measure twice", inputFile);
+            telemetryData.setSurefireClassesOverlapping(telemetryData.getSurefireClassesOverlapping() + 1);
+          }
+        } else if (inputFiles.isEmpty()) {
           telemetryData.setSurefireClassesFailed(telemetryData.getSurefireClassesFailed() + 1);
+        } else {
+          LOGGER.warn("Class name {} resolves to {} files, skipping measure import to avoid ambiguous results", entry.getKey(), inputFiles.size());
+          telemetryData.setSurefireClassesDuplicated(telemetryData.getSurefireClassesDuplicated() + 1);
```
jonas-wielage-sonarsource: Previously we just ignored multiple findings. Why don't we do it now as well and just take the first one? 
Wouldn't that be better in terms of "we tried our best"?

petertrr: Good suggestion, applied
This way we also keep existing behavior

jonas-wielage-sonarsource (approved):

PR #768, opened 2026-08-11
Pin org.sonarqube Gradle plugin to 7.3.1.8318 in kotlin-language-server ruling test

petertrr:
## Summary
- `its/sq-integration` ruling test `SlangRulingTest.test_kotlin_language_server()` has been failing on master since 2026-08-07.
- The sample project `its/sources/kotlin/kotlin-language-server/build.gradle.kts` resolves the `org.sonarqube` Gradle plugin via `version("+")`. Release 7.4.0.8496 regressed a known bug (SCANGRADLE-293: `sonarResolver` task creates a bad implicit dependency link, e.g. with `:shared:processResources`) that 7.1.0.6387 had previously fixed.
- Pin the plugin to the last known-good version, 7.3.1.8318, via a Gradle init script (`its/sq-integration/src/integrationTest/resources/pin-sonarqube-plugin.init.gradle.kts`) passed to the nested Gradle build with `--init-script` in `SlangRulingTest.gradleBuild()`. This is entirely within our own repo and does not touch the `its/sources` submodule.

## Test plan
- [x] Ran `./gradlew --init-script .../pin-sonarqube-plugin.init.gradle.kts build sonar -x test --stacktrace --console=plain -Dsonar.host.url=http://127.0.0.1:9999 -Dsonar.token=dummy -Dsonar.projectKey=test -Dsonar.scanner.skipJreProvisioning=true` directly against `its/sources/kotlin/kotlin-language-server`.
- [x] Confirmed `:sonarResolver`, `:platform:sonarResolver`, `:server:sonarResolver`, `:shared:sonarResolver` all succeed (no "A problem was found with the configuration of task" error).
- [x] Confirmed the build only fails afterwards at `:sonar`, with the expected scanner-bootstrapping error against the dummy server URL.

PR #769, opened 2026-08-11
SONARKT-803 Pin org.sonarqube plugin version in the kotlin-language-server ruling build

matthew-reid-sonarsource:
`its/sources/kotlin/kotlin-language-server` declares `id("org.sonarqube") version "+"`, so every run resolves whatever is newest on the Gradle Plugin Portal. `7.4.0.8496`, published 2026-08-06, declares classpath inputs without task-dependency provenance, so Gradle now fails `:<subproject>:sonarResolver` with an implicit-dependency error against `processResources`. `qa_sq_integration` has been red every night since, on an unchanged commit, blocking all PRs via `promote`.

Pin to `7.3.1.8318` — what the last green run used, and what `build-logic/common` already pins — via a Gradle init script. This also makes the ruling classpath, and therefore the issues asserted here, reproducible rather than dependent on an external release.

Pinned here rather than in the `its/sources` submodule, which is owned by another squad. Drop it once the scanner fix in [SCANGRADLE-440](https://sonarsource.atlassian.net/browse/SCANGRADLE-440) ships.

[SCANGRADLE-440]: https://sonarsource.atlassian.net/browse/SCANGRADLE-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/sq-integration/src/integrationTest/java/org/sonarsource/slang/SlangRulingTest.java
+++ b/its/sq-integration/src/integrationTest/java/org/sonarsource/slang/SlangRulingTest.java
@@ -161,11 +161,16 @@ class SlangRulingTest {
     return moduleDirectory().resolve(Path.of("build", projectKey(project) + "-differences"));
   }
 
+  private static Path pinSonarqubePluginScript() throws IOException {
+    return moduleDirectory().resolve(Path.of("src", "integrationTest", "resources", "pin-sonarqube-plugin.gradle"));
+  }
+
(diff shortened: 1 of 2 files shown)


petertrr (approved):
I stumbled upon this too in #768 :D

matthew-reid-sonarsource:
> I stumbled upon this too in #768 :D

I approved yours instead, it's smaller and CI is already finished running 😄

PR #779, opened 2026-09-01
SONARKT-816 S6619 Fix FP on nullable projections and FP on expression with safe-call operators

lijun-chen-sonarsource:
This PR fixes two FPs:

### FP on nullable projections

The Analysis API represents elements read from out-projected arrays, including nullable `varargs`, as captured types. For `CapturedType(out T?)`, `isMarkedNullable` is false because the captured type itself is not marked with `?`, even though its upper bound allows `null`. S6619 consequently treated nullable elements as non-null and reported their null checks as useless.

Use `KaType.isNullable` so nullability is evaluated semantically, including type bounds. Add regression cases for nullable `varargs`, `requireNotNull`, and explicit out projections.

### FP on expressions with safe-call operators

Due to the upstream Kotlin Analysis API bug KT-66472 (https://youtrack.jetbrains.com/issue/KT-66472), `expressionType` returns the non-null selector type for convention operators pulled into a safe call. For example, it reports `String` for `container?.list[0]` and `Int` for `container?.number++`, although both complete expressions are nullable when container is null. S6619 trusted those types and incorrectly reported subsequent null checks as useless.

As a workaround, we require the corresponding Kotlin compiler diagnostic before reporting comparisons, Elvis operators, safe calls, or non-null assertions. These compiler diagnostics assert that the null check or null-safety operation is useless under the compiler's complete semantics, so S6619 treats them as the source of truth instead of relying on the incorrect Analysis API type. Add regression cases for indexed access and increment operators.

Part of AT-82

----
## Summary by Gitar

- **Bug fixes:**
    - Fixed false positives in `UselessNullCheckCheck` by validating required compiler diagnostics and fixing nullable type checks

### DIFF
--- a/its/ruling/src/integrationTest/resources/expected/kotlin/kotlin/kotlin-S6619.json
+++ b/its/ruling/src/integrationTest/resources/expected/kotlin/kotlin/kotlin-S6619.json
@@ -5,9 +5,6 @@
 "kotlin-kotlin-project:sources/kotlin/kotlin/buildSrc/src/main/kotlin/pill/render.kt": [
 201
 ],
-"kotlin-kotlin-project:sources/kotlin/kotlin/compiler/backend/src/org/jetbrains/kotlin/codegen/coroutines/CoroutineTransformerMethodVisitor.kt": [
-361
-],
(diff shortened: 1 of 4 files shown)


asya-vorobeva (approved):
💯

PR #780, opened 2026-09-02
SONARKT-815 Fix FPs on S100

asya-vorobeva:
Exclude external FFI functions, @Composable UI components and override functions from naming convention checks.

### DIFF
--- a/kotlin-checks-test-sources/src/main/kotlin/checks/BadFunctionNameCheckSample.kt
+++ b/kotlin-checks-test-sources/src/main/kotlin/checks/BadFunctionNameCheckSample.kt
@@ -38,3 +38,13 @@ class BadFunctionNameCheckSample {
     external fun generate_seed() // Compliant - external modifier
 
 }
+
+// Override — name is fixed by the supertype contract, cannot be renamed
+open class BaseWithNonCompliantName {
+    open fun bad_name() {} // Noncompliant {{Rename function "bad_name" to match the regular expression ^[a-zA-Z][a-zA-Z0-9]*$}}
+//           ^^^^^^^^
+}
+
+class DerivedWithOverride : BaseWithNonCompliantName() {
+    override fun bad_name() {} // Compliant - override, name is fixed by supertype
+}
--- a/kotlin-checks-test-sources/src/main/kotlin/checks/BadFunctionNameCheckSample.kt
+++ b/kotlin-checks-test-sources/src/main/kotlin/checks/BadFunctionNameCheckSample.kt
@@ -33,4 +33,8 @@ class BadFunctionNameCheckSample {
     fun `non test backtick name with spaces`() {} // Noncompliant {{Rename function "non test backtick name with spaces" to match the regular expression ^[a-zA-Z][a-zA-Z0-9]*$}}
 //      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+    // External (FFI/JNI) functions — name is owned by the native ABI, renaming would break interop
+    external fun ffi_native_long_function_name() // Compliant - external modifier
+    external fun generate_seed() // Compliant - external modifier
+
 }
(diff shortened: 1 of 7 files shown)


kotlin-checks-test-sources/src/main/kotlin/checks/BadFunctionNameCheckSample.kt
```diff
@@ -33,4 +33,18 @@ class BadFunctionNameCheckSample {
     fun `non test backtick name with spaces`() {} // Noncompliant {{Rename function "non test backtick name with spaces" to match the regular expression ^[a-zA-Z][a-zA-Z0-9]*$}}
 //      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+    // External (FFI/JNI) functions — name is owned by the native ABI, renaming would break interop
+    external fun ffi_native_long_function_name() // Compliant - external modifier
+    external fun generate_seed() // Compliant - external modifier
+
+}
+
+// Override — name is fixed by the supertype contract, cannot be renamed
```
lijun-chen-sonarsource: Having this comment here is slightly confusing. Would it be better to place it at line 47?
Or, make it clear that you're opening a section with multiple classes that participate in the test case.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/BadFunctionNameCheck.kt
```diff
@@ -48,22 +54,52 @@ class BadFunctionNameCheck : AbstractCheck() {
 
     override fun visitNamedFunction(function: KtNamedFunction, kotlinFileContext: KotlinFileContext) {
         val name = function.name ?: /* in case of anonymous functions */ return
-        if (!name.matches(formatRegex) && !isBacktickedTestFunction(function)) {
+        if (function.hasModifier(KtTokens.EXTERNAL_KEYWORD)) return
+        if (function.overrides()) return
+        if (isBacktickedAllowedFunction(function)) return
+        if (name.matches(formatRegex)) return
+        if (function.annotationEntries.any { it.shortName?.asString() == "Composable" }) {
```
lijun-chen-sonarsource: Nitpick: Can `hasAnnotation` be reused here?

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/BadFunctionNameCheck.kt
```diff
@@ -48,22 +54,52 @@ class BadFunctionNameCheck : AbstractCheck() {
...
+            returnsUnit ->
+                name.matches(PASCAL_CASE_REGEX) to PASCAL_CASE_REGEX.pattern
+            else ->
+                name.matches(CAMEL_CASE_REGEX) to CAMEL_CASE_REGEX.pattern
+        }
+
+        if (!follows) {
+            kotlinFileContext.reportIssue(
+                function.nameIdentifier!!,
+                """Rename function "$name" to match the regular expression $expectedPattern"""
```
lijun-chen-sonarsource: IMO it would be confusing to use the same issue message for this case and the normal case.
The user may have configured a `format` and see that the regular expression shown here is not what they've configured and be left confused.
It'd be nice to make it clear that  here we're raising because of non-adherence to JetPack Compose naming conventions. It would also be nice if instead of (or in addition to) showing the regular expression, we directly write "PascalCase", "camelCase" in the issue message. E.g. something like "Rename function "$name" to use PascalCase"

lijun-chen-sonarsource (changes requested):
Overall LGTM!

I left some nitpicks and a request to improve the issue message.

kotlin-checks-test-sources/src/main/kotlin/checks/BadFunctionNameCheckSample.kt
asya-vorobeva: Done

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/BadFunctionNameCheck.kt
asya-vorobeva: Done

lijun-chen-sonarsource (approved):
LGTM!

PR #781, opened 2026-09-02
SONARKT-818 S1192 Introduce several suppression heuristics to reduce noise

lijun-chen-sonarsource:
Reduce noise from S1192 by excluding duplicated string literals in contexts where repetition is usually intentional.

The rule now:

- Skips test files.
- Ignores adjacent literal fragments in concatenations.
- Prevents logging and error-message occurrences from triggering issues:
  - SLF4J, JUL, Log4j 2, and Kotlin Logging
  - Directly thrown exception constructors
  - Kotlin `error`, `require`, and `check`
- Ignores literals inside functions directly annotated with Compose `@Preview`

Message occurrences remain visible when enough ordinary occurrences independently trigger an issue.

Compose suppression is deliberately limited to direct `@Preview` annotations. Custom/multi-preview annotations and separately declared helpers remain out-of-scope now, until we have evidence from user feedback that those should also be suppressed.

Inspired by SONARJAVA-6671

----
## Summary by Gitar

- **Rule updates:**
    - Added suppression heuristics to `StringLiteralDuplicatedCheck` for logging and exception messages
    - Excluded test files from duplicated string literal analysis


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -48,30 +112,47 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
...
                     first,
-                    """Define a constant instead of duplicating this literal "${first.asString()}" $size times.""",
+                    """Define a constant instead of duplicating this literal "$text" $size times.""",
                     secondaryLocations = occurrences.asSequence()
                         .drop(1)
-                        .map { SecondaryLocation(context.textRange(it), "Duplication") }
+                        .map { SecondaryLocation(context.textRange(it.expression), "Duplication") }
                         .toList(),
                     gap = size - 1.0,
                 )
```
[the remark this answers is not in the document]
lijun-chen-sonarsource: Fixed the primary location to use the first triggering occurrence. I
intentionally retained the total occurrence count, gap, and all other
occurrences as secondary locations. Logging and exception occurrences do not
contribute toward reaching the threshold, but once enough ordinary
occurrences trigger an issue, all occurrences may be reported. This matches
the agreed sonar-java-style semantics, so limiting reporting to triggering
occurrences would change the intended behavior.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -80,9 +185,85 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
             node is KtStringTemplateExpression && !node.hasInterpolation() -> sequenceOf(node)
             node is KtAnnotationEntry -> emptySequence()
             node is KtCallExpression && node.isTodoCall() -> emptySequence()
+            node is KtNamedFunction && node.annotationEntries.isNotEmpty() && node.isComposePreview() -> emptySequence()
             else -> node.children.asSequence().flatMap { collectStringTemplates(it) }
         }
 
+    private fun KtNamedFunction.isComposePreview(): Boolean = withKaSession {
+        symbol.annotations.any { it.classId == COMPOSE_PREVIEW_CLASS_ID }
+    }
```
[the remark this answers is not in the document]
lijun-chen-sonarsource: Let's start simple. All the user false positive feedback we got so far was for simple previews.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -80,9 +185,85 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
             node is KtStringTemplateExpression && !node.hasInterpolation() -> sequenceOf(node)
             node is KtAnnotationEntry -> emptySequence()
             node is KtCallExpression && node.isTodoCall() -> emptySequence()
+            node is KtNamedFunction && node.annotationEntries.isNotEmpty() && node.isComposePreview() -> emptySequence()
             else -> node.children.asSequence().flatMap { collectStringTemplates(it) }
         }
 
+    private fun KtNamedFunction.isComposePreview(): Boolean = withKaSession {
+        symbol.annotations.any { it.classId == COMPOSE_PREVIEW_CLASS_ID }
```
asya-vorobeva, line 193: So here we check the same thing as in BadFunctionNameCheck, but with a different way. Let's unify it.
I don't have a strong opinion which way is better, so feel free to select one of two.

kotlin-checks-test-sources/src/main/kotlin/androidx/compose/runtime/Composable.kt
```diff
@@ -0,0 +1,4 @@
+package androidx.compose.runtime
```
asya-vorobeva, line 1: In my tests I've used 
`annotation class Composable
annotation class Preview`

Please unify this part too.

kotlin-checks-test-sources/src/main/kotlin/checks/StringLiteralDuplicatedCheckComposePreviewSample.kt
```diff
@@ -0,0 +1,57 @@
+package checks
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview as ComposePreview
```
asya-vorobeva, line 4: There is also jetbrains' native variant of Preview annotation which is used for similar purposes:
org.jetbrains.compose.ui.tooling.preview.Preview. I'd say that it's compeltely ok for us to simply check for
`hasAnnotation(function, setOf("Preview"))`
to cover both cases.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -37,8 +58,56 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
         private const val DEFAULT_THRESHOLD = 3
         private const val MINIMAL_LITERAL_LENGTH = 5
         private val NO_SEPARATOR_REGEXP = Regex("\\w++")
+        private val COMPOSE_PREVIEW_CLASS_ID = ClassId.fromString("androidx/compose/ui/tooling/preview/Preview")
+
+        private val LOGGING_CALL_MATCHERS = listOf(
```
asya-vorobeva, line 63: Makes sense to add several other popular Kotlin-specific logging solutions:

- io.klogging.logger.Klogger (from KLogging)
- co.touchlab.kermit.Logger (popular solution for Android apps)
- com.github.aakira.napier.Napier (the same)
- io.ktor.client.plugins.logging.Logger (interface for Ktor framework, typically used as wrapper for some other logger solution).

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -48,30 +117,66 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
...
 
-    override fun visitKtFile(file: KtFile, context: KotlinFileContext) {
+    override fun visitKtFile(file: KtFile, context: KotlinFileContext): Unit = withKaSession {
+        if (context.inputFileContext.isTestFile) return
+
         val occurrences = collectStringTemplates(file)
-            .map { it to it.asString() }
-            .filter { (_, text) -> text.length > MINIMAL_LITERAL_LENGTH && !NO_SEPARATOR_REGEXP.matches(text) }
-            .groupBy({ (_, text) -> text }) { it.first }
+            .mapNotNull { expression ->
```
asya-vorobeva, line 168: Let's extract this code to a separate function
```
private fun Expression.toValidLiteralPair(): Pair<String, Expression>? {
    val text = asString()
    return if (text.length > MINIMAL_LITERAL_LENGTH && !NO_SEPARATOR_REGEXP.matches(text)) {
        text to this
    } else {
        null
    }
}
```
and then use it:
`.mapNotNull { it.toValidLiteralPair() }`

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -80,9 +185,85 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
...
+    }
+
+    /** Returns the call when this expression is a direct value argument or the result of a trailing lambda argument. */
+    private fun KtExpression.containingCallArgument(): CallArgument? {
+        val argument = parent as? KtValueArgument
+        val argumentList = argument?.parent as? KtValueArgumentList
+        val directCall = argumentList?.parent as? KtCallExpression
+        if (directCall != null) return CallArgument(directCall, isDirectValueArgument = true)
+
+        val block = parent as? KtBlockExpression ?: return null
```
asya-vorobeva, line 216: I'd suggest to extract it to separate function:
```
private fun KtExpression.isLastInLambdaCall(): KtCallExpression? {
    val block = (parent as? KtBlockExpression)?.takeIf { it.statements.lastOrNull() === this } ?: return null
    val lambdaArgument = (block.parent as? KtFunctionLiteral)
        ?.parent as? KtLambdaExpression
        ?.parent as? KtLambdaArgument ?: return null
    return lambdaArgument.parent as? KtCallExpression
}
```
and then use:
`val lambdaCall = isLastInLambdaCall() ?: return null`

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -48,30 +117,66 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
 
     private fun check(
```
asya-vorobeva, line 118: After all these additions this code is really hard to read. Could we extract some functions in util classes? I mean extension functions added below.
Also makes sense to add comments to explain some decisions.

asya-vorobeva (changes requested):

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
lijun-chen-sonarsource: Thanks for the suggestions! I’ll add support for:

- `co.touchlab.kermit.Logger`: an established, actively maintained Kotlin Multiplatform logger (~1.0k GitHub stars), with clear logging-level APIs (`v/d/i/w/e/a`).
- `com.github.aakira.napier.Napier`: another established KMP/Android logging solution (~1.0k stars), with clear logging intent (`v/d/i/w/e/wtf`).
- `io.ktor.client.plugins.logging.Logger`: Ktor is broadly used (~14.5k stars), and this client-logging interface has the unambiguous `log(message: String)` API.
- `timber.log.Timber`: a very widely used Android logging facade (~10.8k stars), whose severity methods directly represent log messages.
- `android.util.Log`: Android’s built-in logging API. As a rough Kotlin GitHub code-search signal, it has ~977k occurrences (vs. ~139k for Timber), so it is the strongest Android candidate. This will also suppress duplicated tag literals; I think that is an acceptable tradeoff for avoiding noise in the common `Log.d("TAG", "message")` pattern.

I’ll defer KLogging for now. The type is `io.klogging.Klogger` (rather than `io.klogging.logger.Klogger`), and the project is comparatively niche (~250 stars). Supporting it properly would also mean considering its sibling non-coroutine `NoCoLogger` API, rather than adding only a partial match.

lijun-chen-sonarsource: Addressed in 3051594c

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
lijun-chen-sonarsource: Addressed in 04ad33c5

kotlin-checks-test-sources/src/main/kotlin/androidx/compose/runtime/Composable.kt
lijun-chen-sonarsource: Addressed in 04ad33c5

kotlin-checks-test-sources/src/main/kotlin/checks/StringLiteralDuplicatedCheckComposePreviewSample.kt
lijun-chen-sonarsource: Addressed in 04ad33c5

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
lijun-chen-sonarsource: IMO `toValidLiteralPair` is not a good function. It combines two separate tasks: filtering and mapping into pair.

That being said, in 5733eedf I did a big refactor that also touched this code, hoping to make it more clear.
WDYT?

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
lijun-chen-sonarsource: Done in 5733eedf

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
lijun-chen-sonarsource: Yes. Extracted methods in two utility classes in 5733eedf

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/StringLiteralDuplicatedCheck.kt
```diff
@@ -80,9 +186,118 @@ class StringLiteralDuplicatedCheck : AbstractCheck() {
...
+        ),
+        loggingMatcher(
+            "com.github.aakira.napier.Napier",
+            "v", "d", "i", "w", "e", "wtf",
+        ),
+        loggingMatcher("io.ktor.client.plugins.logging.Logger", "log"),
+        loggingMatcher(
+            "timber.log.Timber",
+            "v", "d", "i", "w", "e", "wtf",
+        ),
```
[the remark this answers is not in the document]
lijun-chen-sonarsource: Addressed in 0b73b50f

asya-vorobeva (approved):
💯

PR #786, opened 2026-09-04
SONARKT-819 Authenticate git clones for build-logic/common submodule fetches

teemu-rytilahti-sonarsource:
## Summary
Authenticate the CI submodule fetches for `build-logic/common` so they no longer count against GitHub's anonymous rate limits.

## Details
- Adds a reusable `.github/actions/authenticate-git-clones` composite action (same approach used in `peachee-cloud`): runs `gh auth setup-git` plus `git config --global http.proactiveAuth auto`, and additionally rewrites `git@github.com:` submodule URLs to authenticated HTTPS via `url.insteadOf`.
- Wires it in immediately before the raw `git submodule update --init -- build-logic/common` step, reused via YAML anchor across the `build` and `qa_plugin` jobs. `qa_ruling` already fetches submodules through `actions/checkout`'s own `submodules: true`, which is already authenticated, so it needs no change.
- No behavior change to what gets built or tested — purely CI plumbing.

### DIFF
--- a/.github/actions/authenticate-git-clones/action.yml
+++ b/.github/actions/authenticate-git-clones/action.yml
@@ -1,0 +1,18 @@
+name: Authenticate git clones
+description: Configure git and gh to authenticate clone/fetch requests against GitHub, avoiding anonymous rate limits
+
+inputs:
+  gh_token:
+    description: GitHub token to authenticate with
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      env:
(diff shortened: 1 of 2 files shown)


teemu-rytilahti-sonarsource:
Closing for now — the auth approach needs more work (GitHub doesn't always 401 on requests that actually need auth, so `http.proactiveAuth` is required, but it globally breaks other already-authenticated git operations in the same job). Landing the fix in sonar-text-enterprise first, then reapplying here once finalized.

PR #787, opened 2026-09-04
SONARKT-820 Bump build-logic/common to pick up jackson-databind 2.22.2 fix

teemu-rytilahti-sonarsource:
Bump build-logic/common to pick up jackson-databind 2.22.2 fix.

## Details
- Bumps the `build-logic/common` submodule pointer to `cloud-native-gradle-modules@8eea976`, which raises the minimum jackson-databind version to 2.22.2.
- Fixes the Quality Gate dependency-risk finding caused by the previously pinned commit pulling in an older jackson-databind transitively.


daniel-teuchert-sonarsource (approved):
LGTM!

PR #792, opened 2026-09-07
SONARKT-819 Authenticate git clones for build-logic/common submodule fetches

mstachniuk:
## The problem

`build-logic/common` is fetched by a raw `git submodule update`, which the credentials
`actions/checkout` persists do **not** cover, so the clone goes out anonymously. GitHub
throttles anonymous requests coming from shared runner IPs, which surfaces as:

```
fatal: could not read Username for 'https://github.com'
```

## The fix

A `.github/actions/checkout-build-logic` composite action performs the fetch itself, with
credentials scoped to that single `git` invocation:

```bash
git \
  -c credential.helper= \
  -c 'credential.https://github.com.helper=!f() { if test "$1" = get; then echo username=x-access-token; echo "password=$GH_TOKEN"; fi; }; f' \
  -c http.https://github.com/.proactiveAuth=auto \
  submodule update --init [--depth N] -- build-logic/common
```

Why it is shaped this way:

- **`git -c`, not `git config --global`.** The settings reach the subprocesses that clone the
  submodule (git forwards them through `GIT_CONFIG_PARAMETERS`, so `--recursive` is covered
  too), but they are gone the moment the step ends. Later steps keep using the credentials
  `actions/checkout` persisted, and no other step is affected — in particular
  `ci-github-actions/build-gradle`, which runs its own `git fetch --unshallow` /
  `git fetch origin $GITHUB_BASE_REF`, needs **no** `GH_TOKEN` of its own.
- **Both settings are scoped to `github.com`.** The token is never offered to another host the
  command happens to reach, and no non-github.com remote is forced to authenticate. The empty
  `credential.helper` comes first to reset the helper list, dropping anything inherited from the
  system or global config — notably the `manager` helper that Git for Windows installs, which
  would otherwise be asked first.
- **The helper is an inline shell function that reads `GH_TOKEN`** at call time, so the token
  stays out of the git config and out of `argv`. `gh` is not involved.
- **`proactiveAuth` is required on top of the helper.** GitHub throttles anonymous requests
  without issuing a 401, so lazy authentication never gets an auth challenge to react to and the
  clone fails instead of being retried with credentials. It does not conflict with the
  `http.extraheader` that `actions/checkout` persists, so `git push`/`git fetch` against the
  workspace remote keeps working.

Inputs: `gh-token`, and `depth` (default `'1'`; pass `''` for full history). The action is
byte-identical across the six repos of this initiative.

Call site: the `build` job in `build.yml`, which passes `depth: ''` because it needs the full
history. `qa_ruling` and `qa_sq_integration` fetch submodules through `actions/checkout`'s own
`submodules: true`, which is already authenticated, so they are unchanged.

## ci-github-actions v1 → v2

`get-build-number` and `promote` move to `@v2`. v2 claims build numbers with atomic
`refs/build-number/<N>` git refs instead of a non-atomic repository property, which fixes
concurrent runs (stacked PRs) being handed the same number. Claiming needs `contents: write`, so
the `build` job gets it; `contents: read` returns 403.

## Release workflow

`rule-metadata-pr-labels: "skip-pvf"` is passed to `automated-release.yml`, as the other
analyzers do. This repo was the only one still missing it: `trigger_performance_validation`
already honours a `skip-pvf` label, and the rule-metadata update PR touches no analyzer code, so
performance validation on it is wasted CI time.


petertrr (approved):
LGTM, with the same suggestions an in other repos

PR #793, opened 2026-09-08
SONARKT-822 S6615 Fix FPs when analyzing files with incomplete semantics

lijun-chen-sonarsource:
Suppress unreliable `ASSIGNED_VALUE_IS_NEVER_READ` diagnostics when their
nearest enclosing "region" contains compiler errors.
The compiler errors we observed are due to unresolved symbols, which
realistically occurs when the analysis is performed with partial
semantics.
To avoid suppressing too much and having too many FNs, S6615 findings in
the same file-with-partial-semantics outside the erroneous regions are
preserved.

Add minimal and anonymized regression tests taken from diverse user
feedbacks.

In principle the same kind of compiler-error-induced FPs could arise for the other
two diagnostics (`VARIABLE_NEVER_READ` and `VARIABLE_INITIALIZER_IS_REDUNDANT`)
but I couldn't trigger them and there is no concerning user feedback.
For now I'll apply the heuristic suppression only to `ASSIGNED_VALUE_IS_NEVER_READ`.


francois-mora-sonarsource (changes requested):
`errorRegions` (`UselessAssignmentsCheck.kt:41-42`) is seeded by *any* `ERROR`-severity diagnostic, not just unresolved-symbol ones. An unrelated error in a function (e.g. an argument-type mismatch) suppresses every real `ASSIGNED_VALUE_IS_NEVER_READ` finding in that region — an FN, as already acknowledged in the `mixedResultsInErrorRegion` test. Likely why `corda`/`intellij-rust`/`kotlin` ruling corpora (no classpath) drop to `{}` for S6615.

Suggested fix — narrow the filter:
```kotlin
private val UNRESOLVED_SYMBOL_ERRORS = setOf(
    FirErrors.UNRESOLVED_REFERENCE.name,
    FirErrors.UNRESOLVED_REFERENCE_WRONG_RECEIVER.name,
    FirErrors.UNRESOLVED_IMPORT.name,
    FirErrors.INVISIBLE_REFERENCE.name,
)
val errorRegions = diagnostics
    .filter { it.severity == KaSeverity.ERROR && it.factoryName in UNRESOLVED_SYMBOL_ERRORS }
    .mapNotNullTo(mutableSetOf()) { suppressionRegion(it.psi) }
```
Please rerun ruling on those 3 corpora after narrowing, and confirm the removed findings were genuine FPs.

🤖 Reviewed with Claude Code

lijun-chen-sonarsource:
@francois-mora-sonarsource 

I already sampled the ruling test updates before, but seeing your comment I went through all the issues removed from ruling.
Of all the removed issues, 2 are FNs.

* kotlin-kotlin-project:sources/kotlin/kotlin/libraries/stdlib/test/collections/CollectionTest.kt:442
* kotlin-corda-project:sources/kotlin/corda/testing/test-utils/src/main/kotlin/net/corda/testing/internal/FlowStackSnapshot.kt:37
  Note: FN according to rule, but the useless assignment is arguably meaningful in the test to prove the point of the test.

The rest of genuine FPs that are fixed by this change.

---

To the main comment. I think in general we cannot know with which kind of compiler errors will trigger the FP diagnostics that we're suppressing.  It may happen that other kinds of compiler errors also trigger the FPs. It is a sort of black box (unless we want to invest significant effort understanding, forking/re-implementing this part of the data-flow analysis of the Kotlin compiler).
Also, I'd argue other kinds of compiler errors are not much of a concern since in most cases people submit valid code for Sonar to analyze. In the few cases where they analyze code that is broken, I'd rather have FNs than FPs. 

WDYT?

francois-mora-sonarsource (approved):
Makes sense — thanks for auditing the full removed set. A 1-2 FN cost against corpora with no classpath at all (worst case for this heuristic) is much lower than I assumed, and I agree narrowing to a specific error-diagnostic whitelist would just trade one unproven guess for another given the black-box nature of the underlying compiler behavior. Fine with the current trade-off (prefer rare FNs on already-broken/partial-semantics code over reintroducing FPs). Approving — worth keeping an eye on this if more FN reports come in later.

🤖 Reviewed with Claude Code

PR #799, opened 2026-09-11
SONARKT-825 Set SQS ticket edition and team in release workflow

jonas-wielage-sonarsource:
## Summary
- Add `sqs-ticket-edition: "Community Build & Server"` and `sqs-sqc-ticket-team` (UUID `f1da89c9-3712-4d15-b194-a4b24406e3e4`) to AutomateRelease.yml so the SQS/SQC integration tickets carry the correct Jira Edition and Team fields.

## Test plan
- [ ] Confirm workflow YAML is valid (actionlint/CI)

### DIFF
--- a/.github/workflows/AutomateRelease.yml
+++ b/.github/workflows/AutomateRelease.yml
@@ -80,7 +80,7 @@ jobs:
       rule-metadata-pr-labels: "skip-pvf"
       slack-channel: "squad-security-cloud-notifs"
       create-sli-ticket: true
-      sqs-ticket-edition: "Community Build"
+      sqs-ticket-edition: "Community Build & Server"
       sqs-sqc-ticket-team: "f1da89c9-3712-4d15-b194-a4b24406e3e4"
       runner-environment: sonar-xs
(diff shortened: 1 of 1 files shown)


.github/workflows/AutomateRelease.yml
```diff
@@ -80,6 +80,8 @@ jobs:
       rule-metadata-pr-labels: "skip-pvf"
       slack-channel: "squad-security-cloud-notifs"
       create-sli-ticket: true
+      sqs-ticket-edition: "Community Build"
```
jonas-wielage-sonarsource: Value for this field not clear yet, as we set it to "Community Build" sometimes, sometimes as "Community Build & Server". Will clarify this before merging

jonas-wielage-sonarsource: Changed to `Community Build & Server` as changing kotlin also does affect the Server build. Looking at the documentation I feel it's intended that way.

matthew-reid-sonarsource (approved):

PR #805, opened 2026-09-16
SONARKT-588 S3776: Use a higher threshold of 45 for @Composable functions

asya-vorobeva:
----
## Summary by Gitar

- **Updates:**
    - Increased the cognitive complexity threshold to `45` for `@Composable` functions in `FunctionCognitiveComplexityCheck.kt`


sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/FunctionCognitiveComplexityCheck.kt
```diff
@@ -29,6 +29,7 @@ import java.util.stream.Collectors
 class FunctionCognitiveComplexityCheck : AbstractCheck() {
     companion object {
         private const val DEFAULT_THRESHOLD = 15
+        private const val COMPOSABLE_THRESHOLD = 45
     }
 
     @RuleProperty(key = "threshold",
```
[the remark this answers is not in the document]
asya-vorobeva: Yep, it's ok because Comosable is niche and rather specific case. We can consider this improvement as experimental one. If we'll see improvements in behavior then we can fix rule's description.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/FunctionCognitiveComplexityCheck.kt
```diff
@@ -29,6 +29,7 @@ import java.util.stream.Collectors
 class FunctionCognitiveComplexityCheck : AbstractCheck() {
     companion object {
         private const val DEFAULT_THRESHOLD = 15
+        private const val COMPOSABLE_THRESHOLD = 45
```
lijun-chen-sonarsource, line 32: Just a comment. 45 seems pretty high, but I guess you did your research on SQC and have good reason to put this value.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/FunctionCognitiveComplexityCheck.kt
```diff
@@ -38,16 +39,17 @@ class FunctionCognitiveComplexityCheck : AbstractCheck() {
...
         val nameIdentifier = function.nameIdentifier ?: return
+        val effectiveThreshold = if (function.annotationEntries.any { it.shortName?.asString() == "Composable" }) maxOf(threshold, COMPOSABLE_THRESHOLD) else threshold
         val complexity = CognitiveComplexity(function)
         val value = complexity.value()
-        if (value > threshold) {
+        if (value > effectiveThreshold) {
             context.reportIssue(
                 nameIdentifier,
-                "Refactor this method to reduce its Cognitive Complexity from $value to the $threshold allowed.",
+                "Refactor this method to reduce its Cognitive Complexity from $value to the $effectiveThreshold allowed.",
```
lijun-chen-sonarsource, line 48: I wonder if rspec should be modified to note that Compose functions have a default minimum threshold of 45.
Otherwise it may be surprising for users to see this value and then check the configuration and see a lower value being configured.

lijun-chen-sonarsource (approved):
LGTM!

I just wonder whether rspec should be updated.

sonar-kotlin-checks/src/main/java/org/sonarsource/kotlin/checks/FunctionCognitiveComplexityCheck.kt
asya-vorobeva: I've already clarified in Gitar postcomment that this is experimental fix, we'll see after how the rule's numbers are changed and maybe correlate then. So I prefer not to change Rspec.

